PR #1, opened 2025-02-28
Add MCMS contract

friedemannf:


contracts/packages/mcms/Move.toml
```diff
@@ -0,0 +1,18 @@
+[package]
```
RensR, line 1: can probably remove the `packages` layer and just have `contracts/mcms/Move.toml`

PR #18, opened 2025-03-25
fix decode bytes by adding an offset length check

FelixFan1992:


contracts/ccip/ccip/sources/eth_abi.move
```diff
@@ -233,6 +233,11 @@ module ccip::eth_abi {
         else {
             32 - (length % 32)
         };
+
+        assert!(
```
cfal, line 237: we can just update the above assert?

cfal (approved):

PR #26, opened 2025-03-26
Upgrade aptos CLI to v7

friedemannf:


cfal:
is the CLI already using the new version, or do we need changes in .github?

friedemannf:
> is the CLI already using the new version, or do we need changes in .github?

By default, it'll install the latest version:
https://github.com/smartcontractkit/chainlink-aptos/pull/26/checks#step:3:35

cfal:
let's set it? https://github.com/aptos-labs/actions/blob/main/install-aptos-cli/action.yaml#L9

RensR (approved):

PR #31, opened 2025-03-27
contracts/ccip/ccip/ocr3_base.move: return OCRConfig directly in latest_config_details

cfal:
same as EVM


RensR (approved):

JohnChangUK:
What's the main reason for returning the struct vs primitive values?

cfal:
> What's the main reason for returning the struct vs primitive values?

it makes decoding easier offchain, as other chains also return the struct

PR #39, opened 2025-03-28
steamline uint decoding

RensR:


JohnChangUK (approved):

cfal (changes requested):
i don't agree that this should be merged. could we leave it to be separate functions?

PR #57, opened 2025-03-30
Timelock and batch call functionality, role Auth MCMS

JohnChangUK:
## Description

Support Timelock with MCMS, with batching operations.

- [`execute_batch`, `bypasser_execute_batch`] Batching operation execution
- [`schedule_batch`] Timelock delays per operation
- Role-based access control on functions
- Function blocking/unblocking (function selectors in solidity)
- `Bypasser` role for emergency execution

**Operation Management**
- Scheduling: Operations can be scheduled with configurable delay

## Tests
Tests PR: https://github.com/smartcontractkit/chainlink-aptos/pull/65

Tests will be in a separate PR, purpose of this is to focus on implementation changes.


contracts/mcms/mcms/sources/mcms_account.move
```diff
@@ -91,6 +91,11 @@ module mcms::mcms_account {
         borrow_state().owner
     }
 
+    #[view]
+    public fun is_owner(address: address): bool acquires AccountState {
```
cfal: nit: we've made an effort not to shadow types, use a diff variable name

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -85,68 +162,149 @@ module mcms::mcms {
...
+    }
+
+    #[event]
+    struct RoleRevoked has drop, store {
+        role: u8,
+        account: address,
+        granter: address
+    }
+
+    fun init_module(deployer: &signer) {
```
cfal: nit: publisher for clarity https://github.com/search?q=repo%3Asmartcontractkit%2Fchainlink-aptos+%22fun+init_module%28%22&type=code

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -160,138 +318,121 @@ module mcms::mcms {
                     op_count: 0
                 },
                 root_metadata: RootMetadata {
+                    role,
                     chain_id: 0,
-                    multisig: @mcms,
+                    multisig: signer::address_of(&object_signer),
```
cfal, line 254: this requires additional work to keep track of these object addresses, since it's part of the hash. i am wondering if there's a benefit to this, or if we can't just use the module address as we did previously. the hash preimage already contains the role, so it's not possible for a root from 1 multisig to be re-used on another even without this change.


cc @friedemannf

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -160,138 +318,121 @@ module mcms::mcms {
...
-            metadata.pre_op_count <= metadata.post_op_count,
-            error::invalid_argument(E_WRONG_POST_OP_COUNT)
-        );
-
-        let hashed_leaf: vector<u8> = hash_metadata_leaf(metadata);
-        assert!(
-            verify_merkle_proof(metadata_proof, root, hashed_leaf),
-            error::invalid_argument(E_PROOF_CANNOT_BE_VERIFIED)
+            merkle::verify_merkle_proof(metadata_proof, root, metadata_leaf_hash),
+            E_PROOF_CANNOT_BE_VERIFIED
```
cfal, line 330: what's the reasoning for removing the `error::<type>(..)` wrappers around the error codes? we use them across CCIP as well.

contracts/mcms/mcms/sources/utils/params.move
```diff
@@ -0,0 +1,333 @@
...
+
+    public fun create_call(function: Function, data: vector<u8>): Call {
+        Call { function, data }
+    }
+
+    public fun call_function(call: Call): Function {
+        call.function
+    }
+
+    public fun data(call: Call): vector<u8> {
```
cfal: i don't feel too strongly about this, but requiring these wrappers seem painful, and is a good reason to keep in a single file?

contracts/mcms/mcms/sources/utils/params.move
```diff
@@ -0,0 +1,333 @@
...
+            };
+        };
+
+        // little endian to big endian
+        bcs_bytes.reverse();
+
+        bcs_bytes
+    }
+
+    public fun right_pad_vec(v: &mut vector<u8>, num_bytes: u64) {
```
cfal: this was inline before, can't it still be inline? ditto for other functions

contracts/mcms/mcms/sources/utils/params.move
```diff
@@ -0,0 +1,333 @@
...
+
+    public fun module_name(function: Function): String {
+        function.module_name
+    }
+
+    public fun function_name(function: Function): String {
+        function.function_name
+    }
+
+    public fun deserialize_update_min_delay_params(data: vector<u8>): u64 {
```
cfal: i dont see this used

contracts/mcms/mcms/sources/utils/params.move
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    public fun deserialize_cancel_params(data: vector<u8>): vector<u8> {
+        let stream = bcs_stream::new(data);
+        let id = bcs_stream::deserialize_vector_u8(&mut stream);
+        bcs_stream::assert_is_consumed(&stream);
+        id
+    }
+
+    public fun deserialize_update_delay_params(data: vector<u8>): u64 {
```
cfal: ditto, unused

contracts/mcms/mcms/sources/utils/params.move
```diff
@@ -0,0 +1,333 @@
...
+        let salt = bcs_stream::deserialize_vector_u8(&mut stream);
+        bcs_stream::assert_is_consumed(&stream);
+        (targets, module_names, function_names, datas, predecessor, salt)
+    }
+
+    public fun deserialize_revoke_role_params(data: vector<u8>): (u8, address) {
+        deserialize_grant_role_params(data)
+    }
+
+    public fun deserialize_initialize_params(data: vector<u8>): u64 {
```
cfal: unused

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -432,41 +631,231 @@ module mcms::mcms {
...
                 signer_groups,
                 group_quorums,
                 group_parents,
                 clear_root
             );
+        } else if (function_name_bytes == b"schedule_batch") {
+            assert_bypasser_or_proposer_role(role);
+
+            let (targets, module_names, function_names, datas, predecessor, salt, delay) =
+                params::deserialize_schedule_batch_params(data);
```
cfal: as an overall comment, these functions are used a single time (in this function call). can we keep the deserialization here without a helper function? as a reader, i now have to switch between two files to understand how it's being deserialized. in general, i feel like this is more error prone.

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -160,138 +318,121 @@ module mcms::mcms {
...
         let prev_address = vector[];
         let group_vote_counts: vector<u8> = vector[];
-        right_pad_vec(&mut group_vote_counts, NUM_GROUPS);
+        params::right_pad_vec(&mut group_vote_counts, NUM_GROUPS);
+
         let signatures_len = signatures.length();
         for (i in 0..signatures_len) {
             let signature = signatures[i];
-            let signer_addr = ecdsa_recover_evm_addr(signed_hash, signature);
+            let signer_addr = merkle::ecdsa_recover_evm_addr(signed_hash, signature);
```
cfal: this is not related to signiture verification and not really merkle proofs, it's only used in this module

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -432,41 +631,231 @@ module mcms::mcms {
...
+        let timelock = borrow_timelock();
+        if (!is_bypasser) {
+            assert!(delay >= timelock.min_delay, E_INSUFFICIENT_DELAY);
+        };
+
+        let calls = params::create_calls(targets, module_names, function_names, datas);
+        schedule_batch_internal(caller_address, calls, predecessor, salt, delay);
+    }
+
+    fun schedule_batch_internal(
```
cfal: - nit: inline
- nit: you already borrow the timelock in schedule_batch, and then borrow it again in schedule() in the same function call. why not just borrow &mut Timelock once and pass it into schedule()?

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -432,41 +631,231 @@ module mcms::mcms {
...
+        salt: vector<u8>
+    ) acquires Multisig, MultisigState, Timelock {
+        let calls = params::create_calls(targets, module_names, function_names, datas);
+
+        for (i in 0..calls.length()) {
+            let call = calls[i];
+            let function = params::call_function(call);
+            assert!(
+                !is_function_blocked(
+                    params::target(function),
```
cfal: why not just have is_function_blocked take in a Function instead of extracting each field? you can also create a comparison function in params, eg params::equals(fn1, fn2) and use it in is_function_blocked

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -432,41 +631,231 @@ module mcms::mcms {
...
+        predecessor: vector<u8>,
+        salt: vector<u8>
+    ) acquires Multisig, MultisigState, Timelock {
+        let calls = params::create_calls(targets, module_names, function_names, datas);
+
+        for (i in 0..calls.length()) {
+            let call = calls[i];
+            let function = params::call_function(call);
+            assert!(
+                !is_function_blocked(
```
cfal: this can just be `assert_not_blocked`, that's the only usecase

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -432,41 +631,231 @@ module mcms::mcms {
...
+                &mcms_account::get_signer(), // Must get MCMS signer for `set_config`
+                role,
                 signer_addresses,
                 signer_groups,
                 group_quorums,
                 group_parents,
                 clear_root
             );
+        } else if (function_name_bytes == b"schedule_batch") {
+            assert_bypasser_or_proposer_role(role);
```
cfal: why do we need assert_bypasser_or_proposer_role here when schedule_batch already checks the signer?

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -786,1385 +1450,112 @@ module mcms::mcms {
...
-                bcs_bytes.push_back(0);
-            };
-        };
-
-        // little endian to big endian
-        bcs_bytes.reverse();
-
-        bcs_bytes
+    inline fun borrow(): &MultisigState acquires MultisigState {
+        freeze(borrow_mut())
```
cfal: why is this better than borrow_global directly?

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -432,41 +631,231 @@ module mcms::mcms {
...
+        datas: vector<vector<u8>>,
+        predecessor: vector<u8>,
+        salt: vector<u8>,
+        delay: u64
+    ) acquires MultisigState, Timelock {
+        let caller_address = signer::address_of(caller);
+        let is_bypasser = is_bypasser(caller_address);
+        assert!(
+            is_proposer(caller_address)
+                || is_bypasser
```
cfal: these checks are somewhat meaningless.
in dispatch_to_self, you borrow() the MultisigState, get the Multisig object, get the signer.
here, you borrow() the multisig state, get the Multisig object address, and compare.

`is_bypasser()` and `is_proposer()` can only be true if it matches the object addresses generated by this module, so why make this public at all and add this validation?

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -519,89 +908,110 @@ module mcms::mcms {
...
+                self_signer, object_address, new_owner_address
             );
         } else {
-            abort error::invalid_argument(E_UNKNOWN_MCMS_MODULE_FUNCTION)
+            abort E_UNKNOWN_MCMS_REGISTRY_MODULE_FUNCTION;
+        }
+    }
+
+    fun get_signer(role: Option<u8>): signer acquires Multisig, MultisigState {
+        if (role.is_some()) {
```
cfal: code like this makes me worried as it looks error prone. why not just have get_timelock_role_signer(u8) and get_mcms_signer()?

cfal (commented):
this is great work, thanks John.

although we have put MCMS and timelock in the same class, i think this would be much easier to grok if we put all the timelock code separately (below) the MCMS code so we can see where the boundaries are, with a single init_module where both are initialized.

then we can clearly see that only the timelock calls mcms_registry to dispatch, only mcms dispatches to timelock functions, and so on.

in addition, the signer validation seems unnecessary as mentioned in one of the comments. it's not possible for anyone except a timelock signer to call schedule_batch and related functions now - and the source of truth is always MultisigState+Multisig, so it doesn't seem necessary to have those asserts and comparisons, vs just making those functions private and validating on a u8 role.

i'm also not sure about the dispatch_with_role and dispatch_without_role, imo there should always be a role attached to calls that hit the Timelock functions, and we should be checking that role == bypasser when calling bypasser_execute_batch, not only if option.is_some().

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: We do use `multisig: @mcms` as part of the hash. Friedemann and I discussed yesterday. No need to keep track of the object addresses

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: I don't see a benefit in wrapping this, it makes figuring out what went wrong in our tests much harder as the error code given is now wrapped. 

Removing this allows us to test with specific error codes. Example test below
`#[expected_failure(abort_code = mcms::mcms::E_ALREADY_SEEN_HASH, location = mcms::mcms)]`

With the `error::<type>(..)`, we would have to assert a error code which we would need to decode to hex manually, which makes verification more tedious.

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: Reference: https://github.com/smartcontractkit/chainlink-aptos/pull/57/files#r2023319545

This will be kept as we change `schedule_batch` is private and checks can be removed as we validate the role beforehand. We'll just need to keep the `assert_bypasser_or_proposer_role` before calling schedule_batch

contracts/mcms/mcms/sources/mcms_account.move
JohnChangUK: Will rename to `addr`, cheers

contracts/mcms/mcms/sources/utils/params.move
JohnChangUK: Yes it can be, amended.

contracts/mcms/mcms/sources/utils/params.move
JohnChangUK: No strong preference, can move this back into `mcms` module.

contracts/mcms/mcms/sources/utils/params.move
JohnChangUK: Removed

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: Noted, will remove the helpers and inline them in place.

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: Valid point, the checks can be removed if this function is not public and only callable within the module. This can be private and checks can be removed as we validate the role beforehand. We'll just need to keep the `assert_bypasser_or_proposer_role` before calling `schedule_batch`

Main reason was to for testing, where we can directly pass in the role signer and validate if the caller is the correct role. Instead, tests will just need to create proofs with a different role (canceller), and call execute to verify it cannot call this.

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: No difference, I'll amend to `borrow_global` for explicitness.

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: Removed the `schedule()` and moved this logic into the calling function as its only used once. 

And yes we can pass in the timelock into the functions, removes duplication here.

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: Created the helper func

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: I've abstracted getting both signers into functions to reduce changes of errors here.

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: You're right, have removed this from this file and inlined in `mcms` itself.

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -786,1385 +1030,628 @@ module mcms::mcms {
...
+                }
+            );
         };
+    }
 
-        // little endian to big endian
-        bcs_bytes.reverse();
+    inline fun schedule(
+        timelock: &mut Timelock, id: vector<u8>, delay: u64
+    ) acquires MultisigState {
```
cfal: nit: remove acquires from inline calls, they have no validation and only the parent non-inline functions need it. this function doesn't actually acquire MultisigState.

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -786,1385 +1030,628 @@ module mcms::mcms {
...
+        // Create a single packed vector for all data
+        let packed = vector[];
 
-        assert!(
-            verify_merkle_proof(proof, root, hashed_leaf),
-            1
-        );
+        // Encode the calls array
+        // First encode the length of the calls array
+        packed.append(bcs::to_bytes(&calls.length()));
```
cfal: this no longer replicates keccak256(abi.encode(..)) because bcs packs in little endian. cc @friedemannf

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -786,1385 +1030,628 @@ module mcms::mcms {
...
+        // Then encode each call
+        for (i in 0..calls.length()) {
+            let call = calls.borrow(i);
+            packed.append(bcs::to_bytes(call));
+        };
+
+        packed.append(predecessor);
+        packed.append(salt);
+
+        keccak256(packed)
```
cfal, line 1597: if we're doing our own hashing mechanism since we're using BCS, then we can switch from keccak256 to blake2b for gas savings (and consistency with how we're hashing in keystone/CCIP). cc @friedemannf on if this is preferred/fine for MCMS

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -786,1385 +1030,628 @@ module mcms::mcms {
...
+
+    /// Anyone can call this as it checks if the operation was scheduled by a bypasser or proposer.
+    public entry fun execute_batch(
+        targets: vector<address>,
+        module_names: vector<String>,
+        function_names: vector<String>,
+        datas: vector<vector<u8>>,
+        predecessor: vector<u8>,
+        salt: vector<u8>
+    ) acquires Multisig, MultisigState, Timelock {
```
cfal, line 1200: one pattern we used before is that out outermost function does the acquires, and passes it to all the inline/internal functions. that clearly shows which function is using which resource, and that we are not acquiring multiple times needlessly. can we stick with that approach?

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -786,1385 +1030,628 @@ module mcms::mcms {
...
+        };
+    }
+
+    inline fun after_call(id: vector<u8>) acquires Timelock {
+        assert!(is_operation_ready(id), E_OPERATION_NOT_READY);
+        *borrow_mut_timelock().timestamps.borrow_mut(id) = DONE_TIMESTAMP;
+    }
+
+    /// Anyone can call this as it checks if the operation was scheduled by a bypasser or proposer.
+    public entry fun execute_batch(
```
cfal: a few comments:
- let's prefix all timelock functions with timelock_. this is confusing because we have execute() from the MCMS contract, and execute_batch() from the timelock contract, which has very different behavior.
- let's change execute() to _only_ be able to dispatch to timelock functions. we can eliminate the concepts of "dispatching with roles" and "dispatching without roles". it will always be without roles, as the timelock contract.
- the fact that asserts can be completely skipped by passing in `option::none()` is incredibly worrying and error prone imo. asserts should always occur when the assert function is called, else don't call them at all. please split up the code paths if necessary.

cc @friedemannf

contracts/mcms/mcms/sources/mcms.move
friedemannf: It should be fine for the operation/metadata hashing as that's chain-specific anyway. But not for the tree, as it spans multiple chains. Would we wanna go with two different hashes?

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -160,111 +229,91 @@ module mcms::mcms {
...
-            state.expiring_root_and_op_count.valid_until
-        )
-    }
 
-    #[view]
-    /// Returns the metadata associated with the current Merkle root.
-    public fun get_root_metadata(): RootMetadata acquires MultisigState {
-        borrow_state().root_metadata
+        object::object_from_constructor_ref(constructor_ref)
     }
```
friedemannf, line 263: Can we add these view methods again with a `role` parameter?

friedemannf: Specifically:
- `get_config`
- `get_op_count`
- `get_root`
- `get_root_metadata`

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: Will remove the comment

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: Added them in!

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: Amended from todays discussion

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -696,86 +971,112 @@ module mcms::mcms {
...
-        // abi.encode(root (bytes32), valid_until)
-        let valid_until_bytes = encode_uint(valid_until, 32);
-        assert!(root.length() == 32, error::invalid_argument(E_INVALID_ROOT_LEN)); // root should be 32 bytes
-        let abi_encoded_params = &mut root;
-        abi_encoded_params.append(valid_until_bytes);
+    #[view]
+    public fun get_root(role: u8): vector<u8> acquires MultisigState, Multisig {
+        let multisig = multisig_object(role);
+        borrow_multisig(multisig).expiring_root_and_op_count.root
+    }
```
friedemannf: Lets keep this in line with [EVM](https://github.com/smartcontractkit/ccip-owner-contracts/blob/f3ebba41fd66a7ac68a971996bbde5b3fc33283c/src/ManyChainMultiSig.sol#L508-L508) and return the root and `validUntil`:
```
public fun get_root(): (vector<u8>, u64) acquires MultisigState, Multisig {
```

friedemannf (dismissed):

contracts/mcms/mcms/sources/utils/merkle.move
```diff
@@ -0,0 +1,158 @@
...
+
+        // ECDSA.toEthSignedMessageHash()
+        let eth_msg_prefix = b"\x19Ethereum Signed Message:\n32";
+        let hash = &mut eth_msg_prefix;
+        hash.append(hashed_encoded_params);
+        keccak256(*hash)
+    }
+
+    // computes keccak256(abi.encode(domain_separator, op))
+    public fun hash_op_leaf(domain_separator: vector<u8>, op: Op): vector<u8> {
```
cfal: this is not merkle related either - the op hashing is chain-specific, let's move this back into mcms.move

contracts/mcms/mcms/sources/utils/merkle.move
```diff
@@ -0,0 +1,158 @@
...
+        let pad_amount = 32 - (op.data.length() % 32);
+        params::right_pad_vec(&mut hash_preimage, pad_amount);
+        while (pad_amount > 0) {
+            hash_preimage.push_back(0);
+            pad_amount -= 1;
+        };
+
+        // since we are using this in a merkle tree/proof, hash_preimage should be greater than 64 bytes
+        // to prevent collisions with internal nodes. the above operations already guarantee this so no need to check.
+        keccak256(hash_preimage)
```
cfal: across hash_op_leaf, hash_metadata_leaf, and hash_operation_batch, let's be consistent:
- either little endian or big endian across all fields. i'd suggest just using bcs everywhere if we've already gone down that path
- use blake2b or keccak in all 3

the only place keccak needs to be used is inside that verify_merkle_proof function

contracts/mcms/mcms/sources/utils/merkle.move
```diff
@@ -0,0 +1,158 @@
...
+        abi_encoded_params.append(valid_until_bytes);
+
+        // keccak256(abi_encoded_params)
+        let hashed_encoded_params = keccak256(*abi_encoded_params);
+
+        // ECDSA.toEthSignedMessageHash()
+        let eth_msg_prefix = b"\x19Ethereum Signed Message:\n32";
+        let hash = &mut eth_msg_prefix;
+        hash.append(hashed_encoded_params);
+        keccak256(*hash)
```
cfal: i believe this can also chain-specific and we can use blake2b, since the signing process is already chain-specific - is that true @friedemannf or does that complicate tooling?

contracts/mcms/mcms/sources/utils/merkle.move
```diff
@@ -0,0 +1,158 @@
...
+            hash_preimage.push_back(0);
+            pad_amount -= 1;
+        };
+
+        // since we are using this in a merkle tree/proof, hash_preimage should be greater than 64 bytes
+        // to prevent collisions with internal nodes. the above operations already guarantee this so no need to check.
+        keccak256(hash_preimage)
+    }
+
+    public fun role(op: Op): u8 {
```
cfal: this entire file isn't really merkle related except for `verify_merkle_proof` - could we just merge this back into mcms.move? that would also remove these 50 lines of code since we no longer need getters and you can declare an Op directly

contracts/mcms/mcms/sources/utils/params.move
```diff
@@ -0,0 +1,68 @@
+module mcms::params {
+    use std::bcs;
+
+    const E_CMP_VECTORS_DIFF_LEN: u64 = 0;
+
+    public inline fun encode_uint<T: drop>(input: T, num_bytes: u64): vector<u8> {
```
cfal, line 6: we should be able to remove this if we're using bcs across hashes, same with padding functions below where we can do bcs::to_bytes(var as u32/u64/u128/etc) even if we did want extra padding

contracts/mcms/mcms/sources/utils/params.move
```diff
@@ -0,0 +1,68 @@
...
+            v.reverse();
+            for (i in 0..bytes_to_pad) {
+                v.push_back(0);
+            };
+            v.reverse();
+        };
+    }
+
+    /// compares two vectors of equal length, returns true if a > b, false otherwise.
+    public fun vector_u8_gt(a: &vector<u8>, b: &vector<u8>): bool {
```
cfal, line 34: this is only used in merkle.move, should it be here?

contracts/mcms/mcms/sources/mcms.move
cfal: seems fine to use it everywhere except for merkle hashing - we are already using blake2b in the keystone reports and CCIP code

contracts/mcms/mcms/sources/mcms.move
cfal: i think this would still be a nice cleanup

contracts/mcms/mcms/sources/utils/bcs_stream.move
cfal, line 1: that's cool, i didn't realize we could have subdirs without effect on the import path. i prefer no subdirs, especially if we don't have other files anyway, but don't feel strongly about it.

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: I've made the entry point to the functions do the acquire, and the following function calls omit them, LMK if that looks good

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: Nice, will do

contracts/mcms/mcms/sources/utils/merkle.move
friedemannf: No this needs to stay at keccak256 - `compute_eth_message_hash` has to be consistent across chains as it's used to calculate the payload that is being signed. We take the merkle tree root + the validUntil field and hash it, with the hash being signed and then used across all chains to verify that the signature is correct.

contracts/mcms/mcms/sources/utils/merkle.move
friedemannf: Agree with little-endian/bcs, this would make the tooling implementation easier actually, as we could simply use bcs instead of having to encode manually and leftpad 👍

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -786,1385 +1090,617 @@ module mcms::mcms {
...
+    ): vector<u8> {
+        let packed = vector[];
 
-        assert!(
-            verify_merkle_proof(proof, root, hashed_leaf),
-            1
-        );
+        // Encode the calls array
+        // First encode the length of the calls array
+        packed.append(bcs::to_bytes(&calls.length()));
```
friedemannf: I'm wondering if there's a reason we're not encoding the calls vector directly?
```
        packed.append(bcs::to_bytes(&calls));
```
We're encoding the length as an `u64` here when normally it would be an `uleb128`. We can handle both ways, but might be more idiomatic to serialize the vector "normally"?

JohnChangUK: Good shout, I think we can keep it simple and just encode the calls without any other adjustments.

contracts/mcms/mcms/sources/utils/merkle.move
JohnChangUK: Agreed, will stick to using bcs everywhere, this simplifies the hashing technique too. Also sticking with `keccak256` for all for now.

contracts/mcms/mcms/sources/utils/merkle.move
JohnChangUK: Let's do that

contracts/mcms/mcms/sources/utils/params.move
JohnChangUK: `encode_uint` is used when calling `compute_eth_message_hash` so will keep this for now. The padding functions are for padding vectors, so we still need them. cc @friedemannf

contracts/mcms/mcms/sources/utils/params.move
JohnChangUK: Its also used in `set_root` and `set_config`. Can move this in `mcms.move` if we feel like it should all belong together

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -2,61 +2,141 @@
...
+    const E_UNKNOWN_MCMS_DEPLOYER_MODULE_FUNCTION: u64 = 41;
+    const E_UNKNOWN_MCMS_REGISTRY_MODULE_FUNCTION: u64 = 42;
+    const E_INVALID_PARAMETERS: u64 = 43;
+    const E_INVALID_SIGNATURE_LEN: u64 = 44;
+    const E_INVALID_V_SIGNATURE: u64 = 45;
+    const E_FAILED_ECDSA_RECOVER: u64 = 46;
+    const E_INVALID_MODULE_NAME: u64 = 47;
+    const E_UNKNOWN_MCMS_TIMELOCK_FUNCTION: u64 = 48;
+    const E_INVALID_ROOT_LEN: u64 = 49;
+    const BYPASSER_ROLE: u8 = 0;
```
cfal, line 21: nit: can we move all the errors below struct definitions as we have in another contracts and leave the _ROLE values here?

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -760,1411 +838,978 @@ module mcms::mcms {
...
+            set_config(
+                &mcms_account::get_signer(), // Must get MCMS signer for `set_config`
+                role_param,
+                signer_addresses,
+                signer_groups,
+                group_quorums,
+                group_parents,
+                clear_root
+            );
+        } else if (function_name_bytes == b"timelock_schedule_batch") {
```
cfal: nit: deduplicate this and forward to `dispatch_to_timelock` when the function name begins with `timelock_`?

contracts/mcms/mcms/sources/mcms.move
```diff
@@ -760,1411 +838,978 @@ module mcms::mcms {
...
+        } else {
+            // If role is present, it must be a bypasser (calling from `execute`).
+            let object_meta =
+                mcms_registry::start_dispatch(target, module_name, function_name, data);
+            aptos_framework::dispatchable_fungible_asset::derived_supply(object_meta);
+            mcms_registry::finish_dispatch(target);
+        }
+    }
+
+    inline fun timelock_dispatch_to_self(
```
cfal, line 1277: role is unnecessary, since it's always TIMELOCK_ROLE

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: Yup done

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: removed from this function

JohnChangUK: removed from this function

contracts/mcms/mcms/sources/mcms.move
JohnChangUK: Neat trick, thanks for the tip!

contracts/mcms/mcms/sources/utils/params.move
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public fun vector_starts_with(v: &vector<u8>, prefix: &vector<u8>): bool {
+        let prefix_len = prefix.length();
+
+        if (prefix_len > v.length()) {
+            return false
+        };
+
+        for (i in 0..prefix_len) {
```
cfal: - `inline` modifier for this function or inline the whole function inside timelock_dispatch_to_self since it's only used once?
- this entire function can just be `v.length() >= prefix.length() && v.slice(0, prefix.length()) == prefix`?

JohnChangUK: Yes it can be :). `slice` creates a new array, but this is minimal so will make the change

cfal (dismissed):

cfal (approved):

PR #58, opened 2025-03-31
add nonce manager

RensR:


FelixFan1992 (dismissed):

contracts/ccip/ccip/sources/nonce_manager.move
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    const E_ALREADY_INITIALIZED: u64 = 1;
+
+    #[view]
+    public fun type_and_version(): String {
+        string::utf8(b"NonceManager 1.6.0")
+    }
+
+    fun init_module(publisher: &signer) {
```
FelixFan1992: other modules with state objects usually follow a pattern of
1. register with mcms endpoint in `init_module`
2. have a separate `initialize` function to create the state and transfer the object

is it intentional to deviate from it? `init_module` of nonce manager will be executed together with `init_module` in `state_object`. but i am not familiar with aptos move enough to say if this will create some issues or not.

contracts/ccip/ccip/sources/nonce_manager.move
```diff
@@ -0,0 +1,81 @@
...
+        auth::assert_only_owner(signer::address_of(publisher));
```
cfal: you don't need the auth check in init_module

contracts/ccip/ccip/sources/nonce_manager.move
```diff
@@ -0,0 +1,81 @@
...
+
+        assert!(
```
cfal: you don't need to check for NonceManagerState in init_module, it's impossible for it to exist

contracts/ccip/ccip/sources/nonce_manager.move
```diff
@@ -0,0 +1,106 @@
...
+    friend ccip::onramp;
+    #[test_only]
+    friend ccip::nonce_manager_test;
+
+    struct NonceManagerState has key, store {
+        // dest chain selector -> sender -> nonce
+        outbound_nonces: SmartTable<u64, SmartTable<address, u64>>
+    }
+
+    const E_ALREADY_INITIALIZED: u64 = 1;
```
RensR: to remove

contracts/ccip/ccip/sources/nonce_manager.move
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public fun get_outbound_nonce(
+        dest_chain_selector: u64, sender: address
+    ): u64 acquires NonceManagerState {
+        let state = borrow_state();
+
+        if (!state.outbound_nonces.contains(dest_chain_selector)) {
+            return 0;
+        };
```
JohnChangUK, line 35: Is this section needed if the below defaults to 0 anyway?

JohnChangUK (approved):

contracts/ccip/ccip/sources/nonce_manager.move
RensR: This shortcuts the chain not being there, not the sender! So yes this is required

PR #59, opened 2025-03-31
make onramp config on router

RensR:


contracts/ccip/ccip_router/sources/router.move
```diff
@@ -94,19 +115,31 @@ module ccip_router::router {
...
-        )
+        let state = borrow_state();
+        if (!smart_table::contains(&state.on_ramp_versions, dest_chain_selector)) {
+            abort error::invalid_argument(E_UNKNOWN_CHAIN)
+        };
+
+        let on_ramp_version =
+            smart_table::borrow(&state.on_ramp_versions, dest_chain_selector);
+
+        if (*on_ramp_version == 1) {
```
RensR: on_ramp_version == 1

this works but is a bit random. We could use 16 for 1.6?

contracts/ccip/ccip/sources/onramp_1_6_0.move
```diff
@@ -1,4 +1,4 @@
-module ccip::onramp {
+module ccip::onramp_1_6_0 {
```
cfal: do we need to rename right now? let's leave it as onramp.

contracts/ccip/ccip_router/sources/router.move
```diff
@@ -94,19 +115,31 @@ module ccip_router::router {
...
-            fee_token,
-            fee_token_store,
-            extra_args
-        )
+        let state = borrow_state();
+        if (!smart_table::contains(&state.on_ramp_versions, dest_chain_selector)) {
+            abort error::invalid_argument(E_UNKNOWN_CHAIN)
+        };
+
+        let on_ramp_version =
```
cfal: nit: deref immediately on borrow

contracts/ccip/ccip_router/sources/router.move
cfal: version 1 seems fine to me. and we can leave the onramp module to be onramp, then `onramp2`, etc.

contracts/ccip/ccip_router/sources/router.move
```diff
@@ -116,4 +149,75 @@ module ccip_router::router {
     inline fun borrow_state(): &RouterState {
         borrow_global<RouterState>(@ccip_router)
     }
+
+    inline fun borrow_state_mut(): &mut RouterState {
+        borrow_global_mut<RouterState>(@ccip)
+    }
+
+    // ================================================================
```
cfal, line 169: why do we need this styling?

contracts/ccip/ccip_router/sources/router.move
```diff
@@ -116,4 +149,75 @@ module ccip_router::router {
...
+    ) acquires RouterState {
+        auth::assert_only_owner(signer::address_of(caller));
+
+        let state = borrow_state_mut();
+
+        vector::zip(
+            dest_chain_selectors,
+            ramps_to_use,
+            |dest_chain_selector, uses_blue_on_ramp| {
+                smart_table::upsert(
```
cfal: emit some events, as EVM does when onramps are set

contracts/ccip/ccip_router/sources/router.move
```diff
@@ -116,4 +149,75 @@ module ccip_router::router {
...
+        caller: &signer, dest_chain_selectors: vector<u64>, ramps_to_use: vector<u8>
+    ) acquires RouterState {
+        auth::assert_only_owner(signer::address_of(caller));
+
+        let state = borrow_state_mut();
+
+        vector::zip(
+            dest_chain_selectors,
+            ramps_to_use,
+            |dest_chain_selector, uses_blue_on_ramp| {
```
cfal: what does uses_blue_on_ramp mean? shouldn't this just be called version?

cfal (commented):
let's please keep it as `onramp.move`, and we can use the version numbers in`on_ramp_versions` for future naming.

contracts/ccip/ccip/sources/onramp_1_6_0.move
RensR: We should ensure we don't need to change the name in the future to avoid confusion. If you have another way to do that I'm open to suggestions. Since all onRamps would live in the ccip package the only way to address this is through module names no?

I don't think onRamp1 and onRamp2 are sensible names. We have versions, why not use them?

RensR:
undo naming

use vector mapping

cfal (approved):

PR #60, opened 2025-03-31
Add Aptos support to feeQuoter

RensR:


contracts/ccip/ccip/sources/auth.move
```diff
@@ -75,9 +75,9 @@ module ccip::auth {
         borrow_global_mut<AuthState>(state_object::object_address())
     }
 
-    //
-    // ccip::ownable functions
-    //
+    // ================================================================
```
cfal, line 78: why do we need these styling changes?

contracts/ccip/ccip/sources/fee_quoter.move
```diff
@@ -607,15 +607,18 @@ module ccip::fee_quoter {
         let gas_limit =
             if (chain_family_selector == CHAIN_FAMILY_SELECTOR_EVM) {
                 validate_evm_address(receiver);
-                resolve_evm_gas_limit(dest_chain_config, extra_args)
+                resolve_generic_gas_limit(dest_chain_config, extra_args)
+            } else if (chain_family_selector == CHAIN_FAMILY_SELECTOR_APTOS) {
```
cfal, line 611: i brought this up before and probably outside the immediate scope of this PR, but this path is already so generic/universal - 32 byte address + generic gas limit, it would match many non-EVMs. i wonder if we just shouldn't have a generic/universal family selector.

cfal (commented):
do we need Aptos contracts to have Aptos as a destination family?

contracts/ccip/ccip/sources/auth.move
RensR: Need no, want, yes. This is the header we use for onchain code. There's already many of them in the Aptos code, just like the Solidity code. I was planning on adding more after all features are complete and to reorder some code to group it by functionality. I know the Solidity style guide is called the Solidity style guide but I think quite a few principals are relevant for all onchain code, including the header format that is specified in there

contracts/ccip/ccip/sources/fee_quoter.move
RensR: yeah as discussed offchain we probably want to revisit this at some point. For now I think it's best to merge this PR

cfal (approved):

PR #69, opened 2025-04-01
relayer/chainreader/loop: add loop reader

cfal:


relayer/chainreader/loop/loop_reader.go
```diff
@@ -0,0 +1,240 @@
...
+	if len(readComponents) != 3 {
+		return fmt.Errorf("invalid read identifier: %s", readIdentifier)
+	}
+
+	_, contractName, _ := readComponents[0], readComponents[1], readComponents[2]
+
+	_, ok := a.moduleAddresses[contractName]
+	if !ok {
+		return fmt.Errorf("no such contract: %s", contractName)
+	}
```
sdrug, line 61: We don't need this part, do we? It will fail if the bindings were lost and takes place anyways in the original `GetLatestValue`

relayer/chainreader/chainreader.go
```diff
@@ -507,3 +521,8 @@ func (a *aptosChainReader) Unbind(ctx context.Context, bindings []types.BoundCon
 	}
 	return nil
 }
+
+func (a *aptosChainReader) CreateContractType(readName string, forEncoding bool) (any, error) {
+	// only called when LOOP plugin
+	return &[]byte{}, nil
+}
```
sdrug, line 528: No implementation needed?

relayer/chainreader/loop/loop_reader.go
cfal: this still verifies correctness. this reader is run on the client (core) side, so it makes sure that it was indeed Bind()'d by chainlink-ccip. if this error occurs, that is a bug and not a LOOP connection error.

relayer/chainreader/chainreader.go
cfal: it's always []byte, because we're serializing/deserializing JSON.

PR #73, opened 2025-04-02
Add support for SimpleMaps and Objects to bindgen

friedemannf:


bindings/mcms/mcms_registry/mcms_registry.go
```diff
@@ -362,6 +364,20 @@ func (c mcmsRegistryEncoder) ExecuteCodeObjectTransfer(objectAddress aptos.Accou
 	})
 }
 
+func (c mcmsRegistryEncoder) StartDispatch(callbackAddress aptos.AccountAddress, callbackModuleName string, callbackFunction string, data []byte) (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error) {
```
JohnChangUK, line 367: This is a `(friend)` function, do we need APIs for these?

bindings/bind/stdlib.go
```diff
@@ -14,3 +18,22 @@ func (opt StdOption[T]) Value() *T {
 	}
 	return &opt.Vec[0]
 }
+
+// StdObject is a binding for 0x1::object::Object
+type StdObject struct {
```
JohnChangUK, line 23: Object also has type `T`, but I guess in this case we only care about the address?

friedemannf: Yeah that's right, we can't really do anything with the type parameter, for deserialization is should be enough to just replicate the Object struct: https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/object.move#L131

JohnChangUK (approved):

bindings/mcms/mcms_registry/mcms_registry.go
friedemannf: This is one of the downsides of the current binding generator - as theoretically any function could be callable via mcms, we're generating encoding methods for as many functions as possible. Just in case that it is one of the functions added to the `mcms_entrypoint`. These won't be part of the callable bindings, as only view and entry functions are added there, but will still be added to the encoder

PR #75, opened 2025-04-02
Extract Generalized Write Target

silaslenihan:
Requires:
- https://github.com/smartcontractkit/chainlink-evm/pull/2
- https://github.com/smartcontractkit/chainlink-framework/pull/36


relayer/write_target/aptos/aptos_target_strategy.go
```diff
@@ -0,0 +1,171 @@
+package write_target
```
krebernisak: We should now be able to move these files from `relayer/write_target/aptos` to `relayer/write_target` pkg. Before the aptos subfolder was here just to split the specific from general code.

krebernisak: Also, please rename this file as just `strategy.go` which says enough.

relayer/write_target/aptos/write_target.go
```diff
@@ -44,7 +49,7 @@ func NewAptosWriteTarget(ctx context.Context, chain chain.Chain, lggr logger.Log
 	}
 
 	// Set up a specific Beholder client for the Aptos WT
-	beholder, err := NewAptosWriteTargetMonitor(ctx, lggr)
```
krebernisak, line 47: I think this was named `NewAptosWriteTargetMonitor` for a reason, meaning it's indicating this codepath is Aptos specific. I think it comes down to report decoding, specifically DF and Mercury report decoding for which we use same ABI encoding/decoding for reports on both Ethereum and Aptos, but might change for other networks.

We might want to indicate that with a comment and follow up with a ticket. This can be scoped to a follow-up initiative to extract product-specific code.

krebernisak (approved):
LGTM! Great job splitting this up.

relayer/write_target/aptos/write_target.go
silaslenihan: Added [TODO](https://github.com/smartcontractkit/chainlink-framework/blob/3ed1c84f1f0214908005b746375c5ee8a47255f4/capabilities/writetarget/write_target_monitor.go#L31-L32) in `chainlink-framework` and created [ticket](https://smartcontract-it.atlassian.net/browse/BCFR-1145)

.github/workflows/ci-protobuf.yml
```diff
@@ -1,24 +0,0 @@
...
-    steps:
-      - name: Checkout repo
-        uses: actions/checkout@v4
-
-      - name: Setup buf
-        uses: bufbuild/buf-setup-action@35c243d7f2a909b1d4e40399b348a7fdab27d78d # v1.34.0
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Run buf breaking
```
krebernisak, line 18: Please also move this CI check with the Protobuf definitions, I assume here: https://github.com/smartcontractkit/chainlink-framework/pull/36

krebernisak (approved):

relayer/write_target/types/decode.go
```diff
@@ -0,0 +1,66 @@
...
+	wt_msg "github.com/smartcontractkit/chainlink-evm/pkg/report/pb/platform"
+
+	"github.com/smartcontractkit/chainlink-evm/pkg/report/platform"
+
+	mercury_vX "github.com/smartcontractkit/chainlink-evm/pkg/report/mercury/common"
+	mercury_v3 "github.com/smartcontractkit/chainlink-evm/pkg/report/mercury/v3"
+	mercury_v4 "github.com/smartcontractkit/chainlink-evm/pkg/report/mercury/v4"
+)
+
+func DecodeAsFeedUpdated(m *wt_msg.WriteConfirmed) ([]*registry.FeedUpdated, error) {
```
krebernisak: Please see and follow this comment here: https://github.com/smartcontractkit/chainlink-evm/pull/24#discussion_r2036891432

krebernisak (changes requested):

PR #78, opened 2025-04-04
Fix bindgen - vectors and non-public functions

friedemannf:
This fixes two issues with bindgen:
- non-public functions weren't matched by the treesitter query, updated it to capture those as well and add encoding methods for them. Via mcms virtually any method could be the target of the `mcms_entrypoint`, we therefore need to generate encoding methods for all of them.
- the Move-types for vectors of non-primitive types weren't generated correctly, fixed this by correctly parsing the inner type. I.e `vector<String>` should result in `vector<0x1::string::String>`

Also: removed an outdated replace directive in the go.mod file

### DIFF
--- a/bindings/mcms/mcms/mcms.go
+++ b/bindings/mcms/mcms/mcms.go
@@ -43,9 +43,20 @@ type MCMSEncoder interface {
 	SetRoot(root []byte, validUntil uint64, chainId *big.Int, multisig aptos.AccountAddress, preOpCount uint64, postOpCount uint64, overridePreviousRoot bool, metadataProof [][]byte, signatures [][]byte) (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error)
 	Execute(chainId *big.Int, multisig aptos.AccountAddress, nonce uint64, to aptos.AccountAddress, moduleName string, function string, data []byte, proof [][]byte) (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error)
 	SetConfig(signerAddresses [][]byte, signerGroups []byte, groupQuorums []byte, groupParents []byte, clearRoot bool) (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error)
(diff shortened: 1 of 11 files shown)


bindings/mcms/mcms/mcms.go
```diff
@@ -43,9 +43,20 @@ type MCMSEncoder interface {
...
+	DispatchToSelf(functionNameBytes []byte, data []byte) (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error)
+	DispatchToAccount(functionNameBytes []byte, data []byte) (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error)
+	DispatchToDeployer(functionNameBytes []byte, data []byte) (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error)
+	DispatchToRegistry(functionNameBytes []byte, data []byte) (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error)
+	EcdsaRecoverEvmAddr(ethSignedMessageHash []byte, signature []byte) (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error)
+	ComputeEthMessageHash(root []byte, validUntil uint64) (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error)
+	HashMetadataLeaf(metadata RootMetadata) (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error)
+	HashOpLeaf(op Op) (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error)
+	VerifyMerkleProof(proof [][]byte, root []byte, leaf []byte) (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error)
+	TestRegisterObjectOwnerForNewCodeObject() (bind.ModuleInformation, string, []aptos.TypeTag, [][]byte, error)
```
JohnChangUK, line 56: Do we need the other methods such as schedule_batch, execute_batch?

JohnChangUK (approved):

PR #79, opened 2025-04-07
relay/config: config enhancements

jmank88:
https://smartcontract-it.atlassian.net/browse/CRE-361

Requires:
- https://github.com/smartcontractkit/chainlink-common/pull/1103


relayer/txm/config.go
```diff
@@ -1,34 +1,68 @@
...
+func (c *Config) ValidateConfig() (err error) {
+	if c.ConfirmPollSecs != nil && c.ConfirmPoll != nil {
+		err = errors.Join(err, errors.New("only one of ConfirmPollSecs and Confirm may be set"))
+	} else if c.ConfirmPollSecs != nil {
+		d, err2 := config.NewDuration(time.Duration(*c.ConfirmPollSecs) * time.Second)
+		if err2 != nil {
+			err = errors.Join(err, fmt.Errorf("ConfirmPollSecs must be a positive number: %w", err2))
+		}
+		c.ConfirmPoll = &d
+	}
```
jmank88: I would have rather used methods that return whichever is non-nil, but these field are being directly accessed right now.

relayer/config/config.go
```diff
@@ -42,12 +42,71 @@ type WorkflowConfig struct {
...
 
 type Chain struct {
-	TransactionManager *txm.Config
+	TransactionManager *TxmConfig
 	BalanceMonitor     *monitor.GenericBalanceConfig
 	WriteTargetCap     *write_target.Config
 	Workflow           *WorkflowConfig
 }
 
+type TxmConfig struct {
```
jmank88, line 51: Moved (but aliased) to avoid a cycle.

relayer/config/example.toml
jmank88, line 1: Should this contain any more fields?

relayer/config/docs.toml
jmank88, line 1: We need to document each of these fields. I didn't see existing docs to copy from core or anywhere else in this repo. Is there somewhere else?

PR #81, opened 2025-04-09
Filter QueryKey events by timestamp

sdrug:


cfal (commented):
- that's not how timestamps are specified, it's not a comparator
  ref: https://github.com/smartcontractkit/chainlink-ccip/blob/bee7c32c71cf0aec594c051fef328b4a7281a1fc/pkg/reader/ccip.go#L209
  ref: https://github.com/smartcontractkit/chainlink-common/blob/ed9f50de73222b8a2154366ce9a72c95aa2f97d8/pkg/types/query/query.go#L72
- we need to be able to filter by arbitrary fields inside the event, eg `sequence_number` field, source chain, state for ExecutionStateChanged
  ref: https://github.com/smartcontractkit/chainlink-ccip/blob/bee7c32c71cf0aec594c051fef328b4a7281a1fc/pkg/reader/ccip.go#L357
- should we add something like an `Indexed` ChainReader config param, which when set to true, would cache all events of the specified type? this would speed up searches, we don't need to always do binary searches with 2 calls each. long term, we should switch to a db model
- i'm thinking of adding the timestamp directly into the events - and then we can add a config called `TimestampFieldName`, which would mean we don't need the block RPC call at all
- one good event we can test against as an integration test - MSafe MomentumSafeCreation with over 50000 events: https://api.mainnet.aptoslabs.com/v1/accounts/0xaa90e0d9d16b63ba4a289fb0dc8d1b454058b21c9b5c76864f825d5c1f32582e/events/0xaa90e0d9d16b63ba4a289fb0dc8d1b454058b21c9b5c76864f825d5c1f32582e::creator::MultiSigCreationEvent/events?limit=100
we can test the equality filter by comparing fields like `nonce` and `threshold`

relayer/chainreader/chainreader_local_test.go
```diff
@@ -843,8 +1214,324 @@ func TestLoopChainReaderLocal(t *testing.T) {
...
+			require.NoError(t, err)
+			require.Len(t, seqs, 10)
+			for i := 0; i < len(seqs)-1; i++ {
+				curr := seqs[i].Data.(*SingleValueEvent).SingleUintValue
+				next := seqs[i+1].Data.(*SingleValueEvent).SingleUintValue
+				require.GreaterOrEqual(t, curr, next)
+			}
+		})
+
+		t.Run("ComplexStruct Event", func(t *testing.T) {
```
sdrug: @cfal Test case here

cfal (approved):

PR #82, opened 2025-04-10
pass through DataSource to ChainReader

cfal:


cfal:
squashed into https://github.com/smartcontractkit/chainlink-aptos/pull/81

PR #85, opened 2025-04-10
Switch smart table to big ordered map

JohnChangUK:
Switch smart table to big ordered map

- Includes pagination functionality in `get_all_configured_tokens` as `big_ordered_map` module does not have this.

Aptos Core Commit hash: https://github.com/aptos-labs/aptos-core/commit/837d04bd9bd0dd7156da4252bcfc1d080ed591ec

### DIFF
--- a/bindings/ccip/ccip_test.go
+++ b/bindings/ccip/ccip_test.go
@@ -9,7 +9,7 @@ import (
 
 func TestCompile(t *testing.T) {
 	t.Parallel()
-	output, err := Compile(aptos.AccountOne, aptos.AccountOne, false)
+	output, err := Compile(aptos.AccountOne, aptos.AccountThree, false)
 	require.NoError(t, err)
 	require.NotZero(t, output.Metadata, "Compilation resulted in no metadata")
 	require.NotZero(t, output.Bytecode, "Compilation resulted in no bytecode")
--- a/contracts/ccip/ccip/sources/token_admin_registry.move
+++ b/contracts/ccip/ccip/sources/token_admin_registry.move
@@ -241,12 +241,12 @@ module ccip::token_admin_registry {
 
         let current_key_opt = token_configs.next_key(&start_key);
         if (max_count == 0 || current_key_opt.is_none()) {
-            return (result, start_key, false)
+            return (result, start_key, current_key_opt.is_some())
         };
 
         let current_key = *current_key_opt.borrow();
 
-        for (i in 0..max_count) {
+        for (i in 0..(max_count - 1)) {
             result.push_back(current_key);
 
             let next_key_opt = token_configs.next_key(&current_key);
@@ -255,15 +255,11 @@ module ccip::token_admin_registry {
             };
 
             current_key = *next_key_opt.borrow();
-
-            // Check if there are more tokens after this batch
-            if (i == max_count - 1) {
-                let has_more = token_configs.next_key(&current_key).is_some();
-                return (result, current_key, has_more)
-            }
         };
 
-        (result, current_key, true)
+        // Check if there are more tokens for final max count
+        let has_more = token_configs.next_key(&current_key).is_some();
+        (result, current_key, has_more)
     }
 
     // ================================================================
--- a/contracts/ccip/ccip/sources/token_admin_registry.move
+++ b/contracts/ccip/ccip/sources/token_admin_registry.move
@@ -1021,7 +1021,7 @@ module ccip::token_admin_registry {
     }
 
     #[test(publisher = @ccip)]
-    public fun test_get_all_configured_tokens(publisher: &signer) acquires TokenAdminRegistryState {
+    fun test_get_all_configured_tokens(publisher: &signer) acquires TokenAdminRegistryState {
         state_object::init_module_for_testing(publisher);
         init_module_for_testing(publisher);
 
@@ -1040,7 +1040,7 @@ module ccip::token_admin_registry {
     }
 
     #[test(publisher = @ccip)]
-    public fun test_get_all_configured_tokens_edge_cases(
+    fun test_get_all_configured_tokens_edge_cases(
         publisher: &signer
     ) acquires TokenAdminRegistryState {
         state_object::init_module_for_testing(publisher);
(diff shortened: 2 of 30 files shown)


bindings/ccip/ccip_test.go
```diff
@@ -9,7 +9,7 @@ import (
 
 func TestCompile(t *testing.T) {
 	t.Parallel()
-	output, err := Compile(aptos.AccountOne, aptos.AccountOne, false)
+	output, err := Compile(aptos.AccountOne, aptos.AccountThree, false)
```
JohnChangUK, line 12: Had to change MCMS address as `0x1::bcs_stream` already exists in the Aptos Framework.

contracts/data-feeds/Move.toml
```diff
@@ -14,7 +14,7 @@ owner = "0xcafe"
 platform = "0xbaba"
```
cfal: please leave the data feeds hash

contracts/platform/Move.toml
```diff
@@ -12,6 +12,6 @@ platform = "0x100"
 owner = "0xcafe"
 
 [dependencies]
-AptosFramework = { git = "https://github.com/aptos-labs/aptos-core.git", rev = "34025227eb5367c449bf9dae3cd226fe5c187904", subdir = "aptos-move/framework/aptos-framework" }
+AptosFramework = { git = "https://github.com/aptos-labs/aptos-core.git", rev = "837d04bd9bd0dd7156da4252bcfc1d080ed591ec", subdir = "aptos-move/framework/aptos-framework" }
```
cfal: same with platform

contracts/ccip/ccip/sources/token_admin_registry.move
```diff
@@ -9,7 +9,7 @@ module ccip::token_admin_registry {
     use std::object::{Self, Object, ObjectCore, ExtendRef, TransferRef};
     use std::option::{Self, Option};
     use std::signer;
-    use std::smart_table::{Self, SmartTable};
+    use aptos_std::big_ordered_map::{Self, BigOrderedMap};
```
cfal: nit: keep this std as we do across the package. i know this is differentiated in aptos core code, but it's just 0x1 to use and doesn't really matter

cfal (commented):
please regenerate bindings and make sure smoke test passes

contracts/data-feeds/Move.toml
JohnChangUK: Reverted.

contracts/ccip/ccip/sources/token_admin_registry.move
```diff
@@ -223,16 +220,51 @@ module ccip::token_admin_registry {
     }
 
     #[view]
+    /// Get configured tokens paginated using a start key and limit.
```
FelixFan1992, line 223: maybe we should mention that if someone makes several `get_all_configured_tokens` calls, it's possible certain new keys are missed?

maybe it's already self-explanatory?

JohnChangUK:
> please regenerate bindings and make sure smoke test passes

This is done.

contracts/ccip/ccip/sources/token_admin_registry.move
JohnChangUK: That's only if it's called off chain?

RensR: You should make those calls on a certain block to ensure you have the same state for every call. Maybe worth a note in the comments

contracts/ccip/ccip/sources/token_admin_registry.move
```diff
@@ -223,16 +220,51 @@ module ccip::token_admin_registry {
...
+        let result = vector[];
+
+        let i = 0;
+        while (i < max_count && has_next) {
+            result.push_back(current_key);
+
+            let next_key_opt = token_configs.next_key(&current_key);
+            if (next_key_opt.is_none()) {
+                has_next = false;
+                break;
```
RensR: Do we need has_next if we break anyway? Then we could make the loop a for loop instead of a while as well making it a bit cleaner

contracts/ccip/ccip/sources/token_admin_registry.move
```diff
@@ -223,16 +220,51 @@ module ccip::token_admin_registry {
...
 
-        let state = borrow_state();
-        state.token_configs.keys_paginated(
-            starting_bucket_index, starting_vector_index, max_count
-        )
+        if (max_count == 0) {
+            return (vector[], @0x0, false)
+        };
+
+        let current_key = token_configs.next_key(&start_key);
```
cfal: this doesn't need to be a separate initialization clause, eg
```
let current_key = start_key;
let result = vector[];
let has_next = true;
while (i < max_count) {
  let next_key_opt = token_configs.next_key(&current_key);
  if (next_key_opt.is_none()) {
    has_next = false;
    break;
  }
  let next_key = *next_key_opt.borrow();
  result.push_back(next_key);
  current_key = next_key;
  i++;
}
```

JohnChangUK: Refactored this, PTAL

contracts/ccip/ccip/sources/token_admin_registry.move
JohnChangUK: Yup great point, refactored.

contracts/ccip/ccip/sources/token_admin_registry.move
```diff
@@ -223,16 +220,44 @@ module ccip::token_admin_registry {
...
+
+            let next_key_opt = token_configs.next_key(&current_key);
+            if (next_key_opt.is_none()) {
+                return (result, current_key, false)
+            };
+
+            current_key = *next_key_opt.borrow();
+        };
+
+        (result, current_key, true)
```
cfal: are we sure we don't need some has_next flag? what happens when there are exactly `max_count` items remaining, would this return `true` even when there are no more items?

JohnChangUK: Good point, I've added the check for the last iteration.

cfal (approved):

PR #86, opened 2025-04-11
USDC token pool

RensR:


contracts/ccip/ccip_token_pools/burn_mint_token_pool/sources/burn_mint_token_pool.move
```diff
@@ -29,20 +29,11 @@ module burn_mint_token_pool::burn_mint_token_pool {
         mint_ref: MintRef
     }
 
-    struct RemoteChainConfig has store, drop, copy {
```
RensR, line 36: unused

contracts/ccip/ccip_token_pools/usdc_token_pool/sources/usdc_token_pool.move
```diff
@@ -0,0 +1,657 @@
...
+    // ================================================================
+
+    #[view]
+    public fun type_and_version(): String {
+        string::utf8(b"USDCTokenPool 1.6.0")
+    }
+
+    fun init_module(publisher: &signer) {
+        // register the pool on deployment, because in the case of object code deployment,
+        // this is the only time we have a signer ref to @ccip_usdc_pool.
```
JohnChangUK: is `@ccip_usdc_pool` correct? Or you mean `@usdc_token_pool`?

contracts/ccip/ccip_token_pools/usdc_token_pool/sources/usdc_token_pool.move
```diff
@@ -0,0 +1,657 @@
...
+        token_admin_registry::register_pool(
+            publisher,
+            token_pool_module_name,
+            @local_token,
+            CallbackProof {}
+        );
+
+        // create a resource account to be the owner of the primary FungibleStore we will use.
+        let (store_signer, store_signer_cap) =
+            account::create_resource_account(publisher, STORE_OBJECT_SEED);
```
JohnChangUK, line 100: Why resource account vs creating an object?

cfal: resource account seems reasonable if you don't need object ownership/transfers

contracts/ccip/ccip_token_pools/usdc_token_pool/sources/usdc_token_pool.move
```diff
@@ -0,0 +1,657 @@
...
+    #[view]
+    public fun type_and_version(): String {
+        string::utf8(b"USDCTokenPool 1.6.0")
+    }
+
+    fun init_module(publisher: &signer) {
+        // register the pool on deployment, because in the case of object code deployment,
+        // this is the only time we have a signer ref to @ccip_usdc_pool.
+        assert!(
+            !object::object_exists<Metadata>(@local_token),
```
cfal: this should be `object::object_exists` without the `!`

contracts/ccip/ccip_token_pools/usdc_token_pool/sources/usdc_token_pool.move
```diff
@@ -0,0 +1,657 @@
...
+        // ie. created with primary_fungible_store::create_primary_store_enabled_fungible_asset
+        primary_fungible_store::ensure_primary_store_exists(
+            signer::address_of(&store_signer), metadata
+        );
+
+        move_to(
+            publisher,
+            USDCTokenPoolDeployment {
+                store_signer_cap,
+                ownable_state: ownable::new(publisher, signer::address_of(publisher)),
```
cfal: nit: this can just be `publisher, @usdc_token_pool`

contracts/ccip/ccip_token_pools/usdc_token_pool/sources/usdc_token_pool.move
RensR: For this I mainly copied the existing pool code. 

When you say "don't need object ownership/transfers" do you mean we would not be able to hand over control of the token pool after it's deployed?

cfal (commented):
i guess it's not possible to `git submodule`?

contracts/ccip/ccip_token_pools/usdc_token_pool/sources/usdc_token_pool.move
cfal: no, i mean that you wouldn't be able to decouple the State resource from this module. which is fine, there's no usecase where you would want that

.github/workflows/aptos-movefmt.yml
```diff
@@ -12,8 +12,17 @@ jobs:
           CLI_VERSION: 7.2.0
       - name: Update movefmt
         run: aptos update movefmt --target-version 1.0.7 # Fix version to prevent accidentally breaking the CI
-      - name: Run movefmt
-        working-directory: contracts
+      - name: Run movefmt ccip
```
cfal: this shouldn't be CCIP specific, we're missing `data-feeds`, `tests`, here

we don't have to tackle it in this PR - but can we just have a single script like `contracts/scripts/movefmt-all.sh` which will run it on every dir but skip `vendor/`? we can reuse it in the CI action and do a `git diff` for errors (like we do for the bindings)  @friedemannf

cfal (dismissed):
i can't tell if the USDC functionality is correct, let's merge it for now and test it when we can.

.github/workflows/aptos-movefmt.yml
RensR: Added a GNUmakefile as that's the main way we use these commands in other repo's

friedemannf (approved):

PR #88, opened 2025-04-11
Initial Link Token

JohnChangUK:
## Summary
This PR implements the LINK token, created as a fungible asset. 

There are 2 packages
1. `link-token`
    - This is used for creating tokens without MCMS.
2. `mcms-link-token`
    - Can only be initialized by MCMS only via `mcms_entrypoint`
    - `initialize` and `mint` both called via MCMS

## Notes
Removing the `module_name` assertion in `mcms_registry` as we need to be able to call from another module. We still keep the same address assertion check, therefore this is sufficient.

## Testing
Link token with comprehensive tests.


contracts/link-token/sources/link_token.move
```diff
@@ -0,0 +1,173 @@
...
+    use aptos_framework::object::{Self, Object, ExtendRef};
+    use aptos_framework::primary_fungible_store;
+    use std::option::{Self, Option};
+    use std::signer;
+    use std::string::{Self, String};
+    use mcms::bcs_stream;
+    use mcms::mcms_registry;
+    use std::error;
+
+    const NAME: vector<u8> = b"ChainLink Token";
```
RensR: Would it make sense to have these be params and rename the token to be generic so customers can use it? I know there will be requests from customers for some basic audited ERC-20 counterpart

JohnChangUK: We can do, but I was thinking to keep it consistent with the [LinkToken on EVM](https://github.com/smartcontractkit/LinkToken/blob/master/contracts/v0.6/LinkToken.sol)

This also means we would have to pass in the symbol to `mint`, to specify which FA to mint from. Might be easier to have the symbol in the contract.

RensR: That's the old one, this is actually the latest link token that's deployed on all new chains https://github.com/smartcontractkit/chainlink-evm/blob/develop/contracts/src/v0.8/shared/token/ERC677/LinkToken.sol

It also passes args into a generic one which we use widely for other applications. We will need a token that's not link to test with, and I don't think we want a new program for every token we might deploy correct? 

How is this normally done on Aptos? You cannot expect people to have 5+ token programs just to deploy 5+ tokens right?

contracts/link-token/sources/link_token.move
```diff
@@ -0,0 +1,181 @@
...
+        let management = &ManagementRefs[link_address()];
+        primary_fungible_store::mint(&management.mint_ref, to, amount);
+
+        event::emit(Mint { minter: signer::address_of(minter), to, amount });
+    }
+
+    fun assert_is_owner(caller: &signer) {
+        assert!(
+            signer::address_of(caller)
+                == mcms_registry::get_registered_owner_address(@link),
```
cfal: the token wouldn't be registered before initialization, and MCMS should be optional.

this needs to check object ownership - first
let's assert in init_module that object::is_object(@link), and that `signer::address_of(caller) == object::owner(@link)` or `object::root_owner(@link)`

contracts/link-token/sources/link_token.move
cfal: that seems fine, a single module per token to manage refs. let's not complicate this and create some 1:many token registry.

contracts/link-token/sources/link_token.move
JohnChangUK: Got it, will create a function to assert object ownership. Also any reason we check both the owner and root owner vs calling `object::owns()`? 

We can also instead assert `object::object_exists<PackageRegistry>(addr)` to be more explicit if we insist `@link` must always be a code object.

From `object.move` module

```
   public fun object_exists<T: key>(object: address): bool {
        exists<ObjectCore>(object) && exists_at<T>(object)
    }
```

contracts/link-token/sources/link_token.move
JohnChangUK: That's right, but I think this is fine as we are not creating a generic token template for others to use.

contracts/link-token/sources/link_token.move
```diff
@@ -0,0 +1,450 @@
...
+    const E_NOT_OWNER: u64 = 1;
+    const E_NOT_PUBLISHER: u64 = 2;
+    const E_NOT_ALLOWED_MINTER: u64 = 3;
+    const E_NOT_ALLOWED_BURNER: u64 = 4;
+    const E_TOKEN_NOT_INITIALIZED: u64 = 5;
+    const E_TOKEN_ALREADY_INITIALIZED: u64 = 6;
+
+    /// `publisher` is the code object, deployed through object_code_deployment
+    fun init_module(publisher: &signer) {
+        assert!(object::object_exists<PackageRegistry>(@link), E_NOT_PUBLISHER);
```
JohnChangUK: Change to `is_object()`

contracts/mcms/mcms/sources/mcms_registry.move
```diff
@@ -569,10 +569,6 @@ module mcms::mcms_registry {
             proof_type_info.account_address() == account_address,
             error::invalid_argument(E_PROOF_NOT_AT_ACCOUNT_ADDRESS)
         );
-        assert!(
-            proof_type_info.module_name() == module_name_bytes,
-            error::invalid_argument(E_PROOF_NOT_IN_MODULE)
-        );
```
JohnChangUK, line 575: Relaxing this as we need to add this package from a different module, but same address.

contracts/mcms-link-token/Move.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "MCMSChainlinkToken"
```
cfal: please call this something like MCMSTokenRegistar - this is not an actual token, it just registers the token package on the same object

contracts/link-token/Move.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "ChainlinkToken"
+version = "1.0.0"
+authors = []
+
+[addresses]
+link = "_"
+
+[dev-addresses]
+# Calculated with object_code_deployment code address
+link = "0x8e1c980369b736c76f18e37320e42ba8387afcec18e513e46ef46c7599c02787"
```
cfal: why does this need to be specific?

contracts/link-token/sources/link_token.move
```diff
@@ -0,0 +1,353 @@
...
+        );
+    }
+
+    // ================================================================
+    // |                      Mint/Burn Functions                      |
+    // ================================================================
+
+    public entry fun mint(
+        minter: &signer, to: address, amount: u64
+    ) acquires MetadataRefs, TokenState, TokenData {
```
cfal: i don't see the benefit to 3 resources and 3 borrow's for each mint/burn. can we use the Deployment resource approach from other contracts and put everything in TokenState?

contracts/link-token/sources/link_token.move
```diff
@@ -0,0 +1,353 @@
...
+    const E_TOKEN_ALREADY_INITIALIZED: u64 = 5;
+
+    #[view]
+    public fun type_and_version(): String {
+        string::utf8(b"LinkToken 1.0.0")
+    }
+
+    #[view]
+    public fun token_state_address(): address {
+        object::create_object_address(&@link, TOKEN_STATE_SEED)
```
cfal, line 83: nit: this can call to an inlined token_state_address_internal() with this create_object_address call, and other functions can rely on the same inlined function for a small optimization

contracts/link-token/sources/link_token.move
```diff
@@ -0,0 +1,353 @@
...
+    #[view]
+    public fun is_burner_allowed(burner_address: address): bool acquires TokenState {
+        allowlist::is_allowed(
+            &TokenState[token_state_address()].allowed_burners,
+            burner_address
+        )
+    }
+
+    #[view]
+    public fun token_address(symbol: String): address {
```
cfal: what's the point of this function? symbol is provided on initialize. is metadata() below sufficient?

contracts/link-token/sources/link_token.move
```diff
@@ -0,0 +1,353 @@
...
+        object::create_object_address(&token_state_address(), *symbol.bytes())
+    }
+
+    #[view]
+    public fun metadata(): Object<Metadata> acquires TokenData {
```
cfal: nit: token_metadata?
does returning an Object<Metadata> just return an address on a view call? if not, can this just return `address`?

contracts/link-token/sources/link_token.move
```diff
@@ -0,0 +1,353 @@
...
+            &TokenStateRefs[token_state_address()].extend_ref
+        )
+    }
+
+    public entry fun apply_allowed_minter_updates(
+        caller: &signer,
+        minters_to_remove: vector<address>,
+        minters_to_add: vector<address>
+    ) acquires TokenState {
+        assert_only_owner(signer::address_of(caller));
```
cfal: this acquires two borrows, one in assert_only and then again below when calling apply_allowlist_updates. can we borrow once and pass &TokenState to assert_only_owner? same for burner updates

contracts/link-token/sources/link_token.move
JohnChangUK: This isn't needed, removed

contracts/link-token/sources/link_token.move
JohnChangUK: It returns an object struct, which has an `inner` field with the address. To keep this simple, we can return an `address`

contracts/link-token/Move.toml
JohnChangUK: This is because we create the link token object via `object_code_deployment::publish` [here](https://github.com/smartcontractkit/chainlink-aptos/blob/link-token/contracts/link-token/tests/link_token_tests.move#L23-L40).

We need to know the address because when we call `object_code_deployment::publish`, we know the generated object which we have asserts that this address is an object.

contracts/link-token/sources/link_token.move
JohnChangUK: Good point, added in.

contracts/link-token/sources/link_token.move
JohnChangUK: Amended, great tip

contracts/link-token/sources/link_token.move
JohnChangUK: I agree 3 is too many. I think `MetadataRefs` and `TokenState` should be separate though.

JohnChangUK: Refactored to only have `TokenState` and `TokenMetadataRefs` resources now.

contracts/mcms-link-token/Move.toml
JohnChangUK: Agreed, renamed to `MCMSTokenRegistrar`

contracts/link-token/Move.toml
JohnChangUK: We assert in `init_module` that `@link` is an object.

```
fun init_module(publisher: &signer) {
    assert!(object::is_object(@link), E_NOT_PUBLISHER);
...
}
```

Therefore `@link` must be an object before we call `init_module`. Another way is to create an object beforehand, save the address and use that. But our test case does an end to end test, verifying the code object deployment which I think is nice.

cfal: please take the same approach as i pushed onto the big-ordered-map branch to skip the object check instead: https://github.com/smartcontractkit/chainlink-aptos/blob/big-ordered-map/contracts/ccip/ccip/sources/state_object.move#L31-L38

contracts/mcms-link-token/Move.toml
cfal, line 1: can  you change the folder name to reflect the registrar name change?

JohnChangUK: Naming this MCMSRegistrar instead.

cfal (approved):

PR #120, opened 2025-04-15
bindings/bind/large_packages.go: add missing param for object upgrade

cfal:


friedemannf (approved):
LGTM, great find, ty! 🚀

PR #126, opened 2025-04-17
LinkToken: create an account in init_module

friedemannf:
Create an account in `link_token::init_module` to store EventHandles at

### DIFF
--- a/contracts/link-token/sources/link_token.move
+++ b/contracts/link-token/sources/link_token.move
@@ -1,4 +1,5 @@
 module link::link_token {
+    use std::account;
     use std::event;
     use std::fungible_asset::{Self, BurnRef, Metadata, MintRef, TransferRef};
     use std::object::{Self, ExtendRef, Object, TransferRef as ObjectTransferRef};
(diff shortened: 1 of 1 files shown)


contracts/link-token/sources/link_token.move
```diff
@@ -183,7 +187,7 @@ module link::link_token {
             ownable_state,
             allowed_minters,
             allowed_burners
-        } = move_from(token_state_address);
+        } = move_from<TokenStateDeployment>(token_state_address);
```
JohnChangUK, line 190: Not needed right?

JohnChangUK (dismissed):

contracts/link-token/sources/link_token.move
```diff
@@ -135,6 +136,9 @@ module link::link_token {
         let extend_ref = object::generate_extend_ref(constructor_ref);
         let token_state_signer = &object::generate_signer(constructor_ref);
 
+        // create an Account on the object for event handles.
+        account::create_account_if_does_not_exist(signer::address_of(publisher));
```
cfal: nit: use @link for clarity as we do in other init_module calls

contracts/link-token/sources/link_token.move
friedemannf: Oh my linter was complaining about not being able to infer the type, but it was still compiling - shouldn't hurt to add though?

cfal (approved):

PR #129, opened 2025-04-22
Implemented Hotfix to support LLO reports in the Write Target

silaslenihan:


silaslenihan:
Closed in favor of cleaner https://github.com/smartcontractkit/chainlink-aptos/pull/146

PR #133, opened 2025-04-25
Data Feeds Aptos Registry: Multiple Forwarders + Enable LLO Reports

cawthorne:
Update to Aptos Registry to enable multiple forwarders + enable single value Benchmark Report parsing.

Google Doc for design change:
https://docs.google.com/document/d/1BOsbseisv0jzvCRcpy2xfEQuoMbawz7KheHezYcAzxs/edit?usp=sharing

Note only the `contracts/data-feeds/sources/registry.move` and `contracts/data-feeds/Move.toml` file needs to be changed for the upcoming audit, but to get it to compile locally I needed to duplicate the `platform` folder into `platform` `platform_secondary`.

I intend to remove all changes apart from the `contracts/data-feeds/sources/registry.move` and `contracts/data-feeds/Move.toml` ones before merging this. These are currently present just to help with local testing.

I am open to suggestions on tidying this flow up.

To compile:
`cd contracts/data-feeds/`
`aptos move compile --dev`

To test:
`cd contracts/data-feeds/`
`aptos move test `

To run e2e test for the migration:
`cd contracts/data-feeds/scripts/`
`./test_registry_migration_e2e.sh`

To run e2e test for an initial deployment:
`cd contracts/data-feeds/scripts/`
`./test_registry_deploy_e2e.sh`


cawthorne:
The smoke tests are failing because the build command needs the new named-addresses for platform_b. We can add these in if we choose to keep this directory structure.

contracts/platform_secondary/Move.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "ChainlinkPlatformB"
+version = "1.0.0"
+authors = []
+
+[addresses]
+platform_secondary = "_"
```
krebernisak: There should be a way to reuse the source and define multiple instances of the contract vs. duplicating the whole project just to be able to manifest and reference it with a different label.

I'd expect we can define additional named-addresses in the original `contracts/platform/Move.toml` project.

cc @archseer @cfal, do you guys have suggestions how to best achieve this?

contracts/data-feeds/sources/registry.move
```diff
@@ -442,21 +512,27 @@ module data_feeds::registry {
...
-        // schema is based on first two bytes of the feed id
-        let schema = to_u16be(vector::slice(&report_feed_id, 0, 2));
-
-        let observation_timestamp: u256;
-        let benchmark_price: u256;
-        if (schema == SCHEMA_V3 || schema == SCHEMA_V4) {
-            // offsets are the same for timestamp and benchmark in v3 and v4.
-            observation_timestamp =
-                (to_u32be(vector::slice(&report_data, 3 * 32 - 4, 3 * 32)) as u256);
-            // NOTE: aptos has no signed integer types, so can't parse as i196, this is a raw representation
```
krebernisak, line 455: Why not leave this comment in and add more clarity vs. removing it?

contracts/data-feeds/sources/registry.move
```diff
@@ -393,41 +444,60 @@ module data_feeds::registry {
...
         offset = offset + 32;
 
-        for (i in 0..count) {
+        let feed_ids: vector<vector<u8>> = vector::empty<vector<u8>>();
+        let reports: vector<vector<u8>> = vector::empty<vector<u8>>();
+
+        let is_v03: bool = data_len == count * 13 * 32 + 2 * 32;
+        let is_benchmark: bool = data_len == count * 3 * 32 + 64;
+
+        if (is_v03) {
```
krebernisak, line 538: There's a lot of if/else as part of this decoding, and I'm not sure this is the best way to structure it.

I'd personally prefer flat separate functions that decode different types of reports, with clear documentation and, ideally, a reference back to the original source of the report schema that the function is decoding. That would make the report decoding process a lot more structured, readable, and less prone to bugs. Consider refactoring for clarity.

contracts/data-feeds/sources/registry.move
```diff
@@ -347,6 +364,40 @@ module data_feeds::registry {
         option::none()
     }
 
+    public fun on_report_secondary<T: key>(_meta: object::Object<T>): option::Option<u128> acquires Registry {
```
krebernisak: Please document these two functions and their purpose.

contracts/data-feeds/sources/registry.move
cawthorne: Agreed, I actually did mean to keep it, will re-add. Also I believe the type is wrong, the comment should read i192 I believe.

contracts/platform_secondary/Move.toml
cawthorne: I had no intentions of merging in this `platform_secondary` folder, It was just what I could put together on short notice to enable reviewers to run the tests locally without any additional steps, as we wanted to index on getting people playing specifically with the `registry.move` file asap.

The `platform_secondary` isn't necessary at all for onchain deployment, even with no other changes.

When trying to avoid this, I also thought I should index on not wanting to edit the `contracts/platform/Move.toml`, so stayed away from it to some extent.

I will retry this approach though, before we merge this in, which will be post audit. So we have some lead time there.

Currently priority is `registry.move` for our team, and I don't think the existence of the `platform_secondary` changes the `registry.move` source code for the deploy stage.

contracts/data-feeds/sources/registry.move
cawthorne: I will do, thanks.

My justification at the time was awkwardness around contract upgrades with upgrade policy `compatible`, but actually this is a private function, so the point is moot.

contracts/data-feeds/sources/registry.move
cawthorne: Will do thanks.

contracts/data-feeds/sources/registry.move
cawthorne: I have now split the logic into 2 sub functions. Thanks.

contracts/data-feeds/sources/registry.move
cawthorne: Done, thanks.

contracts/platform_secondary/Move.toml
cawthorne: https://github.com/aptos-labs/aptos-developer-discussions/discussions/694#discussioncomment-13250748

There is a discussion here with someone from Aptos, on this topic.

contracts/data-feeds/sources/registry.move
```diff
@@ -23,6 +23,10 @@ module data_feeds::registry {
         allowed_workflow_names: vector<vector<u8>>
     }
 
+    struct RegistryMigrationStatus has key, store, drop {
```
JohnChangUK, line 26: `store` and `drop` are not needed here

contracts/data-feeds/sources/registry.move
```diff
@@ -170,12 +194,48 @@ module data_feeds::registry {
...
     }
 
+    // This function can be called into to register callbacks in future Forwarders
+    public entry fun register_callbacks(publisher: &signer) acquires Registry {
+        assert!(signer::address_of(publisher) == @data_feeds, ENOT_OWNER);
+
+        assert!(
+            !exists<RegistryMigrationStatus>(get_state_addr()),
+            EALREADY_MIGRATED
+        );
```
JohnChangUK, line 215: We should also be checking if `callback_registered` is `false` too.

contracts/data-feeds/sources/registry.move
```diff
@@ -170,12 +194,48 @@ module data_feeds::registry {
...
+
+        let registry = borrow_global_mut<Registry>(get_state_addr());
+        let extend_ref = &registry.extend_ref;
```
JohnChangUK, line 218: Don't need a separate variable for this

contracts/data-feeds/sources/registry.move
```diff
@@ -170,12 +194,48 @@ module data_feeds::registry {
...
+            );
+        // register to receive platform_secondary::forwarder reports
+        platform_secondary::storage::register(
+            publisher, cb_secondary, new_proof_secondary()
+        );
+
+        move_to(
+            &object_signer,
+            RegistryMigrationStatus { callback_registered: true }
+        );
```
JohnChangUK, line 236: This is duplicated twice, can we add a helper for this

contracts/data-feeds/sources/registry.move
```diff
@@ -690,8 +869,112 @@ module data_feeds::registry {
...
+
+        let (feed_ids, reports) = parse_raw_report(&data);
+        std::debug::print(&feed_ids);
+        std::debug::print(&reports);
```
JohnChangUK, line 910: Remove prints.

contracts/data-feeds/sources/registry.move
cawthorne: I see what you mean. But we never set `callback_registered` to false. In a future upgrade this could be used maybe?

contracts/data-feeds/sources/registry.move
cawthorne: I suppose I thought in future we may want to remove it? Maybe in future upgrades?

JohnChangUK: We can de-structure the resource to remove it. A resource with `key` + `drop` does not make much sense.

contracts/data-feeds/sources/registry.move
JohnChangUK: Yeah in the future, looks good for now :)

contracts/data-feeds/sources/registry.move
cawthorne: fyi, I copied the style from existing `test_parse_raw_report` which was here before.

contracts/data-feeds/sources/registry.move
cawthorne: Thanks, I will change if we get a window to submit another commit for Audit.

contracts/data-feeds/sources/registry.move
cawthorne: Will add if we get a window to submit another commit for Audit.

JohnChangUK (approved):

PR #137, opened 2025-04-28
Added go bindings for data feeds contracts

karen-stepanyan:
Fixed go bind generation script to exclude test functions that have arguments
Added go bindings for data feeds contracts (router, registry)


friedemannf (dismissed):
Nice improvement!

bindings/platform/platform.go
```diff
@@ -0,0 +1,79 @@
...
+		storage:   module_storage.NewStorage(address, client),
+	}
+}
+
+// DeployToObject deploys the Platform contract to a new named object.
+// The resulting address will be calculated using the deployer's account address and the next sequence number
+func DeployToObject(
+	auth aptos.TransactionSigner,
+	client aptos.AptosRpcClient,
+	address aptos.AccountAddress,
```
friedemannf: I think some more meaningful name, like `ownerAddress` would be good here? This is the initial owner of the contracts, right?

bindings/platform/platform.go
```diff
@@ -0,0 +1,79 @@
...
+	platform aptos.AccountAddress,
+) (aptos.AccountAddress, *api.PendingTransaction, Platform, error) {
+	namedAddresses := map[string]aptos.AccountAddress{
+		"owner":    address,
+		"platform": platform,
```
friedemannf: You don't need to manually add the address of the object that the contract will be deployed to here. It will automatically be calculated and added to the named addresses

bindings/data_feeds/data_feeds.go
```diff
@@ -0,0 +1,81 @@
...
+// The resulting address will be calculated using the deployer's account address and the next sequence number
+func DeployToObject(
+	auth aptos.TransactionSigner,
+	client aptos.AptosRpcClient,
+	address aptos.AccountAddress,
+	platform aptos.AccountAddress,
+) (aptos.AccountAddress, *api.PendingTransaction, DataFeeds, error) {
+	namedAddresses := map[string]aptos.AccountAddress{
+		"data_feeds": address,
+		"owner":      address,
```
friedemannf: Same as above, the `data_feeds` named address will be calculated, and `owner` should probably be passed. Right now this would result in the object owning itself, this seems unintentional

friedemannf (approved):

PR #139, opened 2025-04-29
BCFR-1255: WT return success only if we've actually observed acceptance of the report

dhaidashenko:
[Ticket](https://smartcontract-it.atlassian.net/browse/BCFR-1388)
There is a plan to use Workflow’s execution success rate to define the uptime of DF. 

In Aptos' implementation of the Write Target, it returns success once a transaction gets added to the queue. There is still a chance that due to TXM misconfiguration or network congestion, the transaction won’t be included. 

We should return a more accurate transmission status to Workflow DON. Otherwise, the uptime is not accurate.


relayer/write_target/write_target.go
```diff
@@ -55,10 +69,11 @@ type writeTarget struct {
...
 	beholder *monitor.BeholderClient
 
-	cs               commontypes.ChainService
-	cr               commontypes.ContractReader
-	cw               commontypes.ContractWriter
+	cs               chainService
+	cr               contractReader
+	cw               contractWriter
 	configValidateFn func(config ReqConfig) error
+	decodeReport     func(report []byte, metadata capabilities.RequestMetadata) (*platform.Report, error)
```
silaslenihan, line 76: Why make this injected if we're not passing it in anywhere and using a chain-agnostic decode function? Is it just to make it more testable?

GNUmakefile
```diff
@@ -13,4 +13,12 @@ wrappers: ## Generate wrappers for Move contracts.
 
 .PHONY: move-test
 move-test: ## Run the Move tests.
-	contracts/scripts/test.sh
\ No newline at end of file
+	contracts/scripts/test.sh
+
+.PHONY: mockery
+mockery: ## Install mockery.
```
krebernisak: Please notice that this repo has none ad-hoc imperative installation steps anywhere. We use Nix to define our environments in a stable and declarative way.

For example the Protobuf go generator that we used for Beholder messages is declared as part of the default dev env here: https://github.com/smartcontractkit/chainlink-aptos/blob/bb7e84a7ec92510f398a21f2931e2688bcdc154d/shell.nix#L22

You can also declare and use mockery bin in the same way: https://github.com/search?q=repo%3ANixOS%2Fnixpkgs+go-mockery&type=code

At least add `go-mockery` to the `shell.nix` as well.

relayer/write_target/write_target.go
```diff
@@ -45,6 +46,19 @@ const (
 	ContractMethodName_getTransmitter       = "getTransmitter"
 )
 
+type chainService interface {
+	LatestHead(ctx context.Context) (commontypes.Head, error)
+}
+
+type contractReader interface {
```
krebernisak, line 53: These interfaces should be public.

Also, why use anything different that was already designed and used for the generalized write target? We need to merge to a single implementation asap, so it would be best to keep the two instances close as possible.

relayer/write_target/write_target.go
krebernisak: +1 - why do we need to deal with changes to report decoding in this PR?

relayer/write_target/write_target.go
```diff
@@ -230,18 +250,9 @@ func (c *writeTarget) Execute(ctx context.Context, request capabilities.Capabili
 	}
 
 	// Decode the report
-	reportDecoded, err := platform.Decode(inputs.Report)
+	reportDecoded, err := c.decodeReport(inputs.Report, request.Metadata)
```
krebernisak, line 253: The platform report should be a stable high-level report that doesn't change depending on the product, and wraps product report as raw data: https://github.com/smartcontractkit/chainlink-aptos/blob/develop/relayer/report/platform/types.go

If anything, what's volatile here is the underlying product report, not the top level platform report. I don't understand why this general code was restructured and now an injected hook is used.

relayer/write_target/write_target.go
```diff
@@ -438,37 +468,28 @@ func (c *writeTarget) acceptAndConfirmWrite(ctx context.Context, info requestInf
...
-		// Check if the transaction was accepted (included in a chain block, not required to be finalized)
-		// Notice: 'Unconfirmed' is used by TXM to indicate the transaction is not yet included in a block,
-		// while 'Included' (N/A yet) could be used to indicate the transaction is included in a block but not yet finalized.
-		if /* status == commontypes.Included || */ status == commontypes.Finalized {
-			return status, true, nil
-		}
-
-		// false if [Unknown, Pending, Failed, Fatal]
-		return status, false, nil
+	txAccepted, err := c.waitUntilTxReachesFinalStatus(ctx, lggr, txID)
```
krebernisak: `txAccepted` and `txFinal` are two completely different statuses and states.

Reading the new code it's not clear which state are you tracking here.
Looks like the finalized state but not clear.

relayer/write_target/write_target.go
```diff
@@ -517,13 +512,62 @@ func (c *writeTarget) acceptAndConfirmWrite(ctx context.Context, info requestInf
...
+				lggr.Errorw("failed to fetch the transaction status", "txID", txID, "err", err)
+				continue
+			}
+
+			lggr.Debugw("txm - tx status", "txID", txID, "status", status)
+
+			// Check if the transaction was accepted (included in a chain block, not required to be finalized)
+			// Notice: 'Unconfirmed' is used by TXM to indicate the transaction is not yet included in a block,
+			// while 'Included' (N/A yet) could be used to indicate the transaction is included in a block but not yet finalized.
+			// Check if the transaction was accepted (included in a chain block, not required to be finalized)
```
krebernisak: This _"// Check if the transaction was accepted (included in a chain block, not required to be finalized)"_ comment was c/p over twice, and the accepted phase check this comment references was removed and doesn't exist anymore.

Please solidify your design and the implementation here.

relayer/write_target/write_target_test.go
```diff
@@ -95,3 +113,128 @@ func TestNewWriteTargetID(t *testing.T) {
 		})
 	}
 }
+
+func TestWriteTarget_Execute(t *testing.T) {
```
krebernisak: This reads like a bundle of code, there's not much structure to it.

Please make tests readable.

krebernisak (commented):
I don't think this PR improves the codebase, but makes it less structured and readable/clear. It feels like a draft and it's also not clear how this diverges and fits the work on generalized WT in #75 (+ EVM effort).

relayer/write_target/aptos/write_target.go
```diff
@@ -42,7 +42,7 @@ func NewAptosWriteTarget(ctx context.Context, chain chain.Chain, lggr logger.Log
 	}
 
 	// Set up a specific Beholder client for the Aptos WT
-	beholder, err := NewAptosWriteTargetMonitor(ctx, lggr)
+	beholder, err := write_target.NewAptosWriteTargetMonitor(ctx, lggr)
```
krebernisak: Please be careful with the general vs. specific code structure here.

The `NewAptosWriteTargetMonitor` fn was part of the Aptos specific codebase, and it represented a constructor of a Aptos specific WT monitor (using specific processors) from a general WT.

This model now changed as you moved this specific fn to a general codebase pkg, which now hosts the specific `NewAptosWriteTargetMonitor`. We should construct a clear boundary on general vs. specific.

relayer/write_target/write_target.go
dhaidashenko: Yes, it's to make write target more testable. Validation logic requires a valid report. I had 3 options:
1. Create a function that generates a valid report. - not an option since the logic is complex. Including the tests it can double size of the PR
2. Use a hardcoded base64 report provided in one of the tests files. Seems like a bad option since it does not provide a framework for other tests that might depend on report's metadata content.
3. Inject mocked function to decode report.

GNUmakefile
dhaidashenko: Added mockery to the `shell.nix`. Had to bump nixpkgs to a more recent version since go-mockery depends on go's version defined in go.mod. As a result, I was forced to lock golangci-lint since the latest nixpkgs contains version >2.0 that requires a new linter configuration and is not available in our CI.

relayer/write_target/write_target.go
dhaidashenko: I had to define these interfaces in the write_target package to be able to mock them.
There is no need to make them public since `NewAptosWriteTarget` hides all interactions with them.
Making them public pollute the write_target namespace with redundant types that do not provide any value to other pkgs.

relayer/write_target/write_target.go
dhaidashenko: I have not changed the logic and the injection of report decoding is internal to this pkg.
I had to make this change to make write target more testable

relayer/write_target/write_target.go
dhaidashenko: Renamed both to make it more clear.

relayer/write_target/write_target.go
dhaidashenko: Seems like the comment was outdated before my changes. Sorry I've missed that. Removed redundant comments

relayer/write_target/write_target_test.go
dhaidashenko: Tried to make it more readable.

relayer/write_target/aptos/write_target.go
dhaidashenko: Good point. I've moved the NewAptosWriteTargetMonitor pack to the original pkg and injected NoopMonitor.

silaslenihan (approved):
LGTM

relayer/write_target/write_target.go
krebernisak: > redundant types that do not provide any value to other pkgs

This might be a preference thing, but I don't understand why would these types be redundant as these are the main types (hooks) which a client needs to implement to use the `NewWriteTarget` builder.

The builder requires `WriteTargetOpts` which currently hosts requirements as private interfaces. I see no point hiding the declaration of these interfaces and depending on duck typing.

relayer/write_target/write_target.go
krebernisak: This is how the Report can easily be encoded for tests: https://github.com/smartcontractkit/chainlink/blob/develop/core/capabilities/targets/write_target.go#L106

imo the current implementation is a quick fix adding more technical debt. A proper fix is to standardise the report/metadata type and encoding/decoding across these implementations (common, evm, aptos). Please scope out a follow up ticket to resolve this.

Additionally, why is the `capabilities.RequestMetadata` type required to `decodeReport`? This is only bc the code was moved and bundled together quickly and is now coupled in a fn signature.

relayer/write_target/write_target.go
krebernisak: This was obviously slightly confusing to me. `decodeReport` was made configurable, but it's not expected to be configurable (as the OCR report is considered stable atm) but is just exposed to be used in tests to produce a report bc we don't have an easy way to correctly encode a byte[] report, which the WT decodes.

The way to approach it imo:

1. Find an easy way to report encode/decode
    * + help out standardise other implementations if possible
2. An approach like yours, but I'd prefer it more explicit. For example, using a specific private `wt.testingOpts` type which would be set via private `setTestConfiguration` to the WT component by a local test The `wt.testingOpts` type could define args like `skipInputDecoding` and `mockReportProducer`, used by the WT when a specific testing env is set up.

relayer/write_target/write_target.go
```diff
@@ -438,37 +468,28 @@ func (c *writeTarget) acceptAndConfirmWrite(ctx context.Context, info requestInf
...
-		// Check if the transaction was accepted (included in a chain block, not required to be finalized)
-		// Notice: 'Unconfirmed' is used by TXM to indicate the transaction is not yet included in a block,
-		// while 'Included' (N/A yet) could be used to indicate the transaction is included in a block but not yet finalized.
-		if /* status == commontypes.Included || */ status == commontypes.Finalized {
-			return status, true, nil
-		}
-
-		// false if [Unknown, Pending, Failed, Fatal]
-		return status, false, nil
+	txFinalized, err := c.waitTxReachesTerminalStatus(ctx, lggr, txID)
```
krebernisak, line 471: If I'm not mistaken the logic of the WT now changed from:

1. write, run tight loop to check on-chain confirmation, observe/return ASAP
2. write, wait for tx final, check once for on-chain confirmation, observe/return

This is a huge change to the design and functionality of the WT. For example, WT SLA monitoring goes out of the water as we now monitor time to finality, not time to inclusion. The `observed_from_transmitter` flag little less sense then, as observation timings increase.

The *biggest impact* is just the change to synchronous execution of the now send-wait-to-finality capability, which assumes wait times on the order of seconds to minutes in time of congestion, on Ethereum for example, and that the system can handle such back pressure correctly.

Is there a DD where this is outlined and aligned? cc PLG

shell.nix
```diff
@@ -11,10 +11,28 @@
       go_1_23
       gopls
       delve
-      (golangci-lint.override {buildGoModule = buildGo122Module;})
+      # override to lock 1.64 version that is currently used by CI
+      (golangci-lint.overrideAttrs (old: rec {
```
krebernisak, line 15: Nice!

Weird the old version is also not published to nixpkgs as handle, so we need to package it.

relayer/write_target/write_target.go
krebernisak: Thx @dhaidashenko for pointing out this is handled by the underlying CRE cap req/resp model implementation which is asynchronous at this point, and mitigates this concern.

krebernisak (approved):

PR #146, opened 2025-05-07
Implemented Hotfix to support LLO reports in the Write Target

silaslenihan:


relayer/report/llo/types_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// Compare each record field by field.
+	for i := range original {
+		origRecord := original[i]
+		decRecord := (*decoded)[i]
+
+		// Compare FeedID.
+		if origRecord.RemappedID != decRecord.RemappedID {
+			t.Errorf("record %d: mismatched FeedID: expected %x, got %x", i, origRecord.RemappedID, decRecord.RemappedID)
+		}
```
dhaidashenko: Nit: you can make this more compact, while keeping the root cause issues clear:
```suggestion
		assert.Equal(t, origRecord.RemappedID, decRecord.RemappedID, "mismatched FeedID")
```

relayer/monitoring/pb/data-feeds/on-chain/registry/decode.go
```diff
@@ -25,139 +26,122 @@ func DecodeAsFeedUpdated(m *wt_msg.WriteConfirmed) ([]*FeedUpdated, error) {
...
 
-		// Decode the common Mercury report
-		rm, err := mercury_vX.Decode(rf.Data)
-		if err != nil {
-			return nil, fmt.Errorf("failed to decode Mercury report: %w", err)
-		}
+	// HACK: to check if the report is a Mercury report or an LLO report, this will be removed
+	// when the generalized Write Target is completed, as it will allow report schemas to be defined
+	// in the workflow and passed to the Write Target.
+	if dfErr == nil {
```
dhaidashenko, line 34: Nice work making the mapping more compact. 
nit: you could further improve readability by reducing length of `if dfErr == nil {` body, so that you can see the full report processing in a single screen
```
if mercuryErr == nil {
return mercuryReportToFeedUpdate()
}

if lloErr == nil {
return lloReportToFeedUpdate()
}
```
return nil, fmt.Errorf("failed to decode DF Report and LLO Report | Mercury Err: %w, LLO Err: %w", mercuryErr, lloErr)

dhaidashenko (dismissed):
LGTM, left a couple of nits, but feel free to skip them if we get another approval

dhaidashenko (approved):

krebernisak (approved):

PR #154, opened 2025-05-12
Log Write confirmed errors in WT

silaslenihan:


relayer/write_target/write_target.go
```diff
@@ -521,7 +521,10 @@ func (c *writeTarget) acceptAndConfirmWrite(ctx context.Context, info requestInf
 			// Source the transmitter address from the on-chain state
 			info.reportTransmissionState = state
 
-			_ = c.beholder.ProtoEmitter.EmitWithLog(ctx, builder.buildWriteConfirmed(&info, head))
+			err = c.beholder.ProtoEmitter.EmitWithLog(ctx, builder.buildWriteConfirmed(&info, head))
```
dhaidashenko, line 524: Should we log emit error inside EmitWithLog function? I see multiple instances where EmitWithLog is ignored

PR #156, opened 2025-05-12
Add overflow check when calculating token pool local amount

JohnChangUK:
## Description
Add overflow check when calculating token pool local amount

## Tests
Tests added to verify overflow

### DIFF
--- a/contracts/ccip/ccip_token_pools/token_pool/sources/token_pool.move
+++ b/contracts/ccip/ccip_token_pools/token_pool/sources/token_pool.move
@@ -15,6 +15,8 @@ module ccip_token_pool::token_pool {
     use ccip_token_pool::token_pool_rate_limiter;
 
     const STORE_OBJECT_SEED: vector<u8> = b"CCIPTokenPool";
+    const MAX_U256: u256 =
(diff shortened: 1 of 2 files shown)


contracts/ccip/ccip_token_pools/token_pool/sources/token_pool.move
```diff
@@ -637,101 +654,3 @@ module ccip_token_pool::token_pool {
         token_pool_rate_limiter::destroy_rate_limiter(rate_limiter_config);
     }
 }
-
-#[test_only]
-module ccip_token_pool::token_pool_test {
```
JohnChangUK, line 642: Tests moved to `token_pool_test.move`.

RensR (approved):

contracts/ccip/ccip_token_pools/token_pool/sources/token_pool.move
```diff
@@ -527,7 +530,6 @@ module ccip_token_pool::token_pool {
     fun calculate_local_amount_internal(
```
FelixFan1992, line 530: should we just check if the value is no larger than `18446744073709551615`?
but in `calculate_local_amount`, it will always compare against it to make sure it fits into u64.

PR #157, opened 2025-05-13
Add validation checks on function inputs

JohnChangUK:
## Description

Add validation checks on function inputs.

## Fee quoter
When calling `decode_generic_extra_args`, matched [EVM implementation](https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/ccip/FeeQuoter.sol#L957-L964) to 
1. `if length == 0`, return default values
2. [Extra check] Add assertion that length <= 4, to prevent slice reverting TX


contracts/ccip/ccip/sources/fee_quoter.move
```diff
@@ -417,6 +418,23 @@ module ccip::fee_quoter {
         let token_transfer_fee_configs =
             state.token_transfer_fee_configs.borrow_mut(dest_chain_selector);
 
+        remove_tokens.for_each_ref(
```
JohnChangUK: Remove tokens before adding tokens, to prevent tokens being removed after added

RensR: This would be a deviation from EVM. Generally, we use arguments in the same order as we get them, meaning if we take a list of removes before adds, we will do removes before adds. In this case, in both EVM and here we take adds before removes and should apply them in that order. 

Generally we do removes first, but let's remain consistent with the existing code

contracts/ccip/ccip/sources/fee_quoter.move
```diff
@@ -763,7 +769,7 @@ module ccip::fee_quoter {
             dest_chain_config.default_tx_gas_limit as u256
```
RensR: probably OK to remove this now that you moved it to `decode_generic_extra_args`. This appears to be wrong anyway as it skips the OOO check

contracts/ccip/ccip_router/sources/router.move
```diff
@@ -239,12 +239,17 @@ module ccip_router::router {
...
     #[view]
     /// Returns the onRamp versions for the given destination chains.
+    /// For chain selectors that do not exist, an empty vector is returned.
     public fun get_on_ramp_versions(
         dest_chain_selectors: vector<u64>
     ): vector<vector<u8>> acquires RouterState {
         let state = borrow_state();
         dest_chain_selectors.map((|dest_chain_selector| {
-            *state.on_ramp_versions.borrow(dest_chain_selector)
+            if (state.on_ramp_versions.contains(dest_chain_selector)) {
```
RensR: Cleaner to use default 

```
        dest_chain_selectors.map((|dest_chain_selector| {
            *state.on_ramp_versions.borrow_with_default(dest_chain_selector, &vector[])
        }))
```

contracts/ccip/ccip_router/sources/router.move
```diff
@@ -259,6 +264,11 @@ module ccip_router::router {
         dest_chain_selectors: vector<u64>,
         on_ramp_versions: vector<vector<u8>>
     ) acquires RouterState {
+        assert!(
+            dest_chain_selectors.length() == on_ramp_versions.length(),
+            error::invalid_argument(E_INVALID_ON_RAMP_VERSION)
```
RensR: Maybe a better error?

contracts/ccip/ccip/sources/fee_quoter.move
```diff
@@ -627,6 +628,11 @@ module ccip::fee_quoter {
             };
 
         let fee_token_price = get_token_price_internal(state, fee_token);
+        assert!(
+            fee_token_price.value > 0,
+            error::invalid_state(E_ZERO_TOKEN_PRICE)
```
RensR, line 633: A price of 0 should be valid. EVM just checks for staleness and returns a fresher price if it has one. If it doesn't it returns the value, which can be 0

contracts/ccip/ccip/sources/fee_quoter.move
JohnChangUK: Cool, will revert back to original

contracts/ccip/ccip/sources/fee_quoter.move
JohnChangUK: Yes good spot :), removed

contracts/ccip/ccip_router/sources/router.move
JohnChangUK: Changed to `E_SET_ON_RAMP_VERSIONS_MISMATCH`

RensR (approved):

PR #161, opened 2025-05-13
Add bindings for LINK token

friedemannf:
This adds bindings for the `link-token` package.

Also, add methods to compile and deploy the `mcms-registrar` package on top of an already deployed LINK token.

This adds support for serializing/deserializing `0x1::option::Option` types in bcs util, required as the `link_token::initialize` function takes an option parameter


bindings/link-token/link_token.go
```diff
@@ -0,0 +1,122 @@
...
+	// Compile using CLI
+	return compile.CompilePackage(contracts.LinkToken, namedAddresses)
+}
+
+func Bind(address aptos.AccountAddress, client aptos.AptosRpcClient) LinkToken {
+	return LinkTokenContact{
+		address:   address,
+		allowlist: module_allowlist.NewAllowlist(address, client),
+		ownable:   module_ownable.NewOwnable(address, client),
+	}
```
rodrigombsoares, line 68: Needs linkToken here

friedemannf: Thanks, fixed!

contracts/mcms-registrar/sources/mcms_token_registrar.move
```diff
@@ -20,7 +20,7 @@ module link::mcms_token_registrar {
 
         if (@mcms_register_entrypoints == @0x1) {
             mcms_registry::register_entrypoint(
-                publisher, string::utf8(b"mcms_token_registrar"), McmsCallback {}
+                publisher, string::utf8(b"link_token"), McmsCallback {}
```
friedemannf, line 23: Changing the module name that the `mcms_token_registrar` registers with with mcms to match what the bindings will generate

JohnChangUK: This would fail on registration? ` !registration.callback_modules.contains(module_name_bytes)`

JohnChangUK: Nice spot

JohnChangUK (dismissed):

contracts/mcms/mcms/sources/mcms_registry.move
```diff
@@ -595,7 +595,9 @@ module mcms::mcms_registry {
 
         let callback_function_info =
             function_info::new_function_info(
-                account, module_name, string::utf8(b"mcms_entrypoint")
+                account,
+                string::utf8(proof_type_info.module_name()),
+                string::utf8(b"mcms_entrypoint")
```
friedemannf, line 604: Changing the module name that's used to construct the function_info from the passed string to `T`'s module name.

Reason being that we might register the entrypoints from another module that what they're actually calling:
e.g. with `mcms_token_registrar` which registers the entrypoints for `link_token`. When calling methods on `link_token` via MCMS we reasonably expect to call them with `link_token` being the module name but they'd currently have to be `mcms_token_registrar` instead

JohnChangUK (dismissed):

sdrug (approved):

PR #164, opened 2025-05-14
PLEX-1438 Allow adding custom tags to WT ID

dhaidashenko:
In the current implementation, the WT ID is programmatically generated from the chain ID and version, and there is no way to change it via configuration.

To test active-active, we need to enable two writer DONs with the same capability.  Unfortunately, this requires two separate capabilities registries used by two workflow DONs.  

Deploying a new capabilities registry and migrating the Workflow DON requires effort similar to making the WT ID taggable (e.g. write_aptos-testnet:region_a@1.0.3), but making it taggable gives us more flexibility.

### DIFF
--- a/relayer/write_target/config.go
+++ b/relayer/write_target/config.go
@@ -8,6 +8,7 @@ import (
 
 // Config defines the write target component configuration.
 type Config struct {
+	Tag                 string // allows modifying WT ID e.g. write_aptos-testnet:{{.Tag}}@1.0.3
 	ConfirmerPollPeriod config.Duration
 	ConfirmerTimeout    config.Duration
 }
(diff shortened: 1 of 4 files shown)


silaslenihan (approved):

relayer/write_target/config.go
```diff
@@ -8,6 +8,7 @@ import (
 
 // Config defines the write target component configuration.
 type Config struct {
+	Tag                 string // allows modifying WT ID e.g. write_aptos-testnet:{{.Tag}}@1.0.3
```
krebernisak, line 11: Not sure why use `:` as a delimiter here, instead of just `-`,
as `testnet` is also a "tag", right?

krebernisak (approved):

relayer/write_target/config.go
krebernisak: This needs some higher level standardisation from the core Keystone team now that tagging/namespacing capabilities is introduced. Good enough for now.

dhaidashenko: There is a standard defined by a [regex](https://github.com/smartcontractkit/chainlink-common/blob/main/pkg/capabilities/capabilities.go#L343) that validates ID. E.g. {name}:{label1_key}_{labe1_value}:{label2_key}_{label2_value}@{version}.

PR #166, opened 2025-05-15
Update latest head to seconds

silaslenihan:


relayer/write_target/write_target.go
```diff
@@ -278,6 +278,7 @@ func (c *writeTarget) Execute(ctx context.Context, request capabilities.Capabili
 		msg := builder.buildWriteError(info, 0, "failed to fetch the latest head", err.Error())
 		return capabilities.CapabilityResponse{}, c.asEmittedError(ctx, msg)
 	}
+	head.Timestamp = head.Timestamp * 1000 // Convert to milliseconds
```
krebernisak: imo `head.Timestamp` should either host ms or s - maybe introduce a different local var representing time in ms?

silaslenihan: I moved this conversion to the message builders to conserve the `head.Timestamp` val

krebernisak: Thanks!

krebernisak (approved):

PR #192, opened 2025-05-23
contracts/ccip/ccip_token_pools: events to store address

cfal:


contracts/ccip/ccip_token_pools/token_pool/sources/token_pool.move
```diff
@@ -115,40 +115,25 @@ module ccip_token_pool::token_pool {
     /// This function should be called from the init_module function to ensure the events
     /// are created on the correct object.
     public fun initialize(
```
JohnChangUK, line 117: Could probably name this `new` since this is just creating a `TokenPoolState`.

JohnChangUK (approved):

contracts/ccip/ccip_token_pools/token_pool/sources/token_pool.move
cfal: +1, `new` makes more sense. i'll leave it as-is to avoid additional changes though.

PR #193, opened 2025-05-26
Synthetic Events + Various fixes

sdrug:


relayer/chainreader/chainreader.go
```diff
@@ -29,35 +32,40 @@ type aptosChainReader struct {
 	types.UnimplementedContractReader
 
 	logger  logger.Logger
-	config  ChainReaderConfig
-	dbStore *DBStore
-	starter commonutils.StartStopOnce
+	config  config.ChainReaderConfig
+	dbStore *db.DBStore
+
+	starter             commonutils.StartStopOnce
```
faisal-chainlink, line 38: This can be replaced with `services.StateMachine`. Should be a drop-in replacement AFAIK.

relayer/chainreader/chainreader.go
```diff
@@ -77,123 +85,35 @@ func (a *aptosChainReader) HealthReport() map[string]error {
...
+			a.logger.Infow("AptosChainReader started event polling", "interval", a.config.EventSyncInterval)
 
-		if len(records) > 0 {
-			if err := a.dbStore.InsertEvents(ctx, records); err != nil {
-				return fmt.Errorf("syncEvent: failed to insert new events: %w", err)
-			}
+			var syncTxCtx context.Context
+			syncTxCtx, a.txSyncCancelFunc = context.WithCancel(ctx)
+			go a.startTxPolling(syncTxCtx)
+			a.logger.Infow("AptosChainReader started transaction polling", "interval", a.config.TxSyncInterval)
```
faisal-chainlink, line 98: This approach is much cleaner! Love it

relayer/chainreader/chainreader_testnet_test.go
```diff
@@ -19,18 +30,141 @@ func TestChainReaderTestnet(t *testing.T) {
...
+							"message_id": {
+								NewName: "MessageId",
+							},
+							"message_hash": {
+								NewName: "MessageHash",
+							},
+							"state": {
+								NewName: "State",
+							},
+						},
```
faisal-chainlink: If all renames are changing between pascal case and underscored, we could possibly write a helper that does this recursively on the returned DB results without keep a config in the events config

relayer/chainreader/chainreader_testnet_test.go
```diff
@@ -19,18 +30,141 @@ func TestChainReaderTestnet(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	err = chainReader.Start(ctx)
+	require.NoError(t, err)
+	defer chainReader.Close()
+
+	logger.Infow("ChainReader started, will run for 1 minute to observe event and tx polling logs")
+
+	time.Sleep(2 * time.Minute)
```
faisal-chainlink: Perhaps we could use `t.Eventually` instead of waiting for a set time

relayer/chainreader/event_poller.go
```diff
@@ -0,0 +1,202 @@
...
+					a.logger.Errorw("SyncAllEvents: error syncing event", "module", moduleKey, "event", eventKey, "error", err)
+				} else {
+					successCount++
+				}
+			}
+		}
+	}
+
+	if errorCount > 0 {
+		a.logger.Errorw("SyncAllEvents: completed with errors", "successCount", successCount, "errorCount", errorCount, "lastError", lastErr)
```
faisal-chainlink, line 196: What happens after we emit this error? I assume we'll depend on monitoring to catch these instances (if any).

faisal-chainlink (dismissed):
Looks good, left some small nitpick comments

faisal-chainlink:
@sdrug I didn't run the tests locally so just a heads up

relayer/chainreader/chainreader_testnet_test.go
sdrug: That's not only about snake case -> camel case conversion. The renaming also help mapping the fields in the original event to the fields the consumer expects, e.g. `source_chain_selector` in the event to `SourceChain` in the receiver. We have to keep this config because we have no control over how the fields are named in the CCIP plugin.

relayer/chainreader/chainreader_testnet_test.go
sdrug: I added this test to showcase the minimal configuration required for successful testing. I'm going to remove it before merging.

relayer/chainreader/event_poller.go
sdrug: For now it's just a heads-up in the logs. Sure we'd want to generate an error message here when we plug-in Beholder. For now monitoring takes place externally.

friedemannf (approved):

relayer/chainreader/event_poller.go
```diff
@@ -0,0 +1,202 @@
...
+				if err := a.dbStore.InsertEvents(ctx, batchRecords); err != nil {
+					return fmt.Errorf("syncEvent: failed to insert batch of events: %w", err)
+				}
+
+				totalProcessed += len(batchRecords)
+				a.logger.Debugw("syncEvent: saved batch of events",
+					"batch_count", len(batchRecords),
+					"total_processed", totalProcessed,
+					"handle", eventHandle)
+			}
```
jporter1979, line 131: This code adds events to the database in batches. However, if a single event in the batch fails to be added to the database the entire batch will not be added. Although, the polling mechanism will retry the batch in the next round, if the erroneous event persists, it will prevent the current and all future batches from being added to the database. Instead of returning immediately on any error, process each event individually if the batch insert fails, and skip those that truly cannot be inserted. Accordingly, the code can be updated as follows:

``` golang
if len(batchRecords) > 0 {
    err := a.dbStore.InsertEvents(ctx, batchRecords)
    if err != nil {
        a.lggr.Errorw("syncEvent: batch insert failed, falling back to per-event insert", "error", err)
        for i, record := range batchRecords {
            if err := a.dbStore.InsertEvents(ctx, []db.EventRecord{record}); err != nil {
                a.lggr.Errorw("syncEvent: failed to insert single event, skipping", "event", record, "error", err)
                continue
            }
            totalProcessed++
            // Advance latestOffset for each successfully inserted event
            latestOffset = *record.EventOffset + 1
        }
    } else {
        totalProcessed += len(batchRecords)
        latestOffset = *batchRecords[len(batchRecords)-1].EventOffset + 1
    }
    a.lggr.Debugw("syncEvent: saved batch of events",
        "batch_count", len(batchRecords),
        "total_processed", totalProcessed,
        "handle", eventHandle)
}
```
This will avoid any erroneous events from preventing an entire batch from being added to the database. The error will be logged and the `latestOffset` value will be updated to the last successfully added event.

PR #196, opened 2025-05-28
Add platform_secondary package + add to deployment CI

cawthorne:
This patch adds a new package `platform_secondary`, which is needed to deploy a second Forwarder for the upcoming Data Feeds Registry upgrade.

Full Data Feeds Registry upgrade PR:
https://github.com/smartcontractkit/chainlink-aptos/pull/133

### DIFF
--- a/bindings/platform_secondary.go
+++ b/bindings/platform_secondary.go
@@ -1,0 +1,76 @@
+package platform_secondary
+
+import (
+	"github.com/aptos-labs/aptos-go-sdk"
+	"github.com/aptos-labs/aptos-go-sdk/api"
+	"github.com/smartcontractkit/chainlink-aptos/bindings/bind"
+	"github.com/smartcontractkit/chainlink-aptos/bindings/compile"
+	module_forwarder "github.com/smartcontractkit/chainlink-aptos/bindings/platform_secondary/forwarder"
+	module_storage "github.com/smartcontractkit/chainlink-aptos/bindings/platform_secondary/storage"
+	"github.com/smartcontractkit/chainlink-aptos/contracts"
+)
+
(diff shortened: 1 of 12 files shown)


silaslenihan (commented):
Can you also generate bindings for the secondary contract?

integration-tests/deploy/devnet.go
```diff
@@ -127,7 +127,7 @@ func (d *Deployer) DeployPlatform() error {
 		"--package-dir=/contracts/platform",
 		"--address-name=platform",
 		"--named-addresses",
-		fmt.Sprintf("platform=%s,owner=%s", DEVNET_ACC, DEVNET_ACC),
```
dhaidashenko, line 130: This does not seem to be correct. We pass two values into sprintf, but there is only one %s

cawthorne: Outdated.

silaslenihan:
I believe we also need to duplicate [this file](https://github.com/smartcontractkit/chainlink-aptos/blob/484cfa7ccddf318706d5e5f32f150c0571b7bf52/bindings/platform/platform.go) to deploy the contract, is that correct?

silaslenihan (approved):

dhaidashenko (approved):

karen-stepanyan (approved):

bindings/platform_secondary.go
```diff
@@ -0,0 +1,76 @@
+package platform_secondary
```
rodrigombsoares, line 1: This file shouldn't be in the root of `/bindings`, there was something wrong in the paths, I think.
eg: if you generate another binding like this there'll be a conflict of `package` name.

PR #204, opened 2025-06-03
Zero/validation/range checks

JohnChangUK:
## Desc
Add zero/validation/range checks


contracts/ccip/ccip/sources/eth_abi.move
```diff
@@ -93,7 +94,12 @@ module ccip::eth_abi {
         out: &mut vector<u8>, value: vector<u8>
```
RensR, line 108: I believe this is unused so maybe we just delete it?

contracts/ccip/ccip/sources/eth_abi.move
```diff
@@ -53,11 +53,12 @@ module ccip::eth_abi {
...
     ) {
         assert!(value.length() <= 32, E_INVALID_BYTES32_LENGTH);
+
+        out.append(value);
```
RensR, line 71: Does this not change the padding to be right padded which is incorrect?

We should have an encoding function for bytes (right padded) and for non-bytes (left padded)

contracts/ccip/ccip/sources/eth_abi.move
JohnChangUK: `encode_bytes` is used by onramp, offramp and rmn_remote.

contracts/ccip/ccip/sources/eth_abi.move
JohnChangUK: Yup you're right, added left padding for (address, int) types and right padding for bytes.

contracts/ccip/ccip_offramp/sources/offramp.move
```diff
@@ -1135,18 +1150,6 @@ module ccip_offramp::offramp {
         )
     }
 
-    inline fun deserialize_execution_reports(reports_bytes: vector<u8>):
```
JohnChangUK, line 1138: Function not used

RensR (approved):

PR #205, opened 2025-06-04
Add custom token registrars to TokenAdminRegistry

friedemannf:
We need to be able toe deploy token pools independently from tokens, i.e. to separate objects.
This currently leads to issues when deploying token pools via mcms, as each object will have a different resource account owner, failing both current checks that:
1. The token is owned by the same (root) owner as the token pool
2. CCIP is owned by the same (root) owner as the token pool

To mitigate this, we're adding an additional (optional) mapping of registrars that can be used to allow an arbitrary address to register a token pool.
In practice this allows us to:
1. Calculate the mcms owner address of the to-be-deployed token pool
2. Add this owner address as an additional registrar for the token that a token pool should be deployed to, by calling `set_token_registrar(local_token, owner_address)`
3. Use mcms to deploy the token pool, this will now pass `assert_can_register`

### DIFF
--- a/contracts/ccip/ccip/sources/token_admin_registry.move
+++ b/contracts/ccip/ccip/sources/token_admin_registry.move
@@ -12,7 +12,6 @@ module ccip::token_admin_registry {
     use std::big_ordered_map::{Self, BigOrderedMap};
     use std::string::{Self, String};
     use std::type_info::{Self, TypeInfo};
-    use aptos_std::ordered_map::OrderedMap;
 
     use ccip::auth;
     use ccip::state_object;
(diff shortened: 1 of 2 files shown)


contracts/ccip/ccip/sources/token_admin_registry.move
```diff
@@ -472,6 +491,30 @@ module ccip::token_admin_registry {
         );
     }
 
+    public entry fun set_token_registrar(
+        caller: &signer, local_token: address, token_registrar: address
+    ) acquires TokenAdminRegistryState {
+        auth::assert_only_owner(signer::address_of(caller));
+
+        let state = borrow_state_mut();
+        state.token_registrars.add(local_token, token_registrar)
```
JohnChangUK: Maybe add an event after setting. We could also use `upsert` instead if we wanted to replace the same key without aborting

JohnChangUK (dismissed):

contracts/ccip/ccip/sources/token_admin_registry.move
friedemannf: Added, thanks!

JohnChangUK (approved):

PR #213, opened 2025-06-07
Add checks, assertions, minor refactoring

JohnChangUK:
## Desc
Add length checks, assertions, minor refactoring

### DIFF
--- a/contracts/ccip/ccip_offramp/sources/ocr3_base.move
+++ b/contracts/ccip/ccip_offramp/sources/ocr3_base.move
@@ -83,6 +83,7 @@ module ccip_offramp::ocr3_base {
     const E_NON_UNIQUE_SIGNATURES: u64 = 18;
     const E_INVALID_SIGNATURE: u64 = 19;
     const E_ZERO_ADDRESS_NOT_ALLOWED: u64 = 20;
+    const E_INVALID_SIGNATURE_LENGTH: u64 = 21;
 
     public fun new(event_account: &signer): OCR3BaseState {
         OCR3BaseState {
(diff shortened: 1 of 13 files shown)


contracts/ccip/ccip_router/sources/router.move
```diff
@@ -190,6 +192,11 @@ module ccip_router::router {
         fee_token_store: address,
         extra_args: vector<u8>
     ): vector<u8> acquires RouterState {
+        assert!(
```
RensR: I believe we didn't add it here because it's done on the first call into the onRamp so this doesn't add much security. The important one to check is the message execution on the Router, but since that doesn't exist through the router imo this contract doesn't need RMN.


If we assume there's some exploit from the Router, RMN wouldn't really guard against anything for Aptos, as the worst (most permissioned) thing it can do is call ccip_send on the onramp. As long as that has the RMN check we should be good rigth?


Let's add a comment stating that if you agree

JohnChangUK: Removed, cheers

contracts/ccip/ccip_offramp/sources/ocr3_base.move
```diff
@@ -346,6 +352,12 @@ module ccip_offramp::ocr3_base {
         let seen = bit_vector::new(signers.length());
         signatures.for_each_ref(
             |signature_bytes| {
+                let signature_bytes: &vector<u8> = signature_bytes;
```
cfal, line 355: is there a compiler issue that requires this? interesting that it's necessary when adding `.length()` but `.slice()` below works

cfal (approved):

contracts/ccip/ccip_offramp/sources/ocr3_base.move
JohnChangUK: It is. When I add the `.length()` check, it's when compilation fails, 
```
error: constraint `fun self.length(_):integer` incompatible with `fun self.slice(_,integer,integer):_`
    ┌─ /Users/johnchang/chainlink/code/chainlink-aptos/contracts/ccip/ccip_offramp/sources/ocr3_base.move:363:25
    │
363 │                         signature_bytes.slice(0, 32)
```

Example in Aptos: https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/aptos_coin.move#L115-L118

PR #216, opened 2025-06-10
Data Feeds Add Revert To Get Reports

cawthorne:
New features:

-Adding a revert to `data_feeds::router::get_reports`, as we are proposing to disbale this function.
-Removing the writing of the report_data as part of the Feed structure.
-Zeroing it out existing report values
-Adding a read event in the Router to help onchain tracking

IMPORTANT:
This PR must be reviewed as a contract **_upgrade_** with upgrade policy 'upgrade', to an already deployed Data Feeds Package.

### DIFF
--- a/contracts/data-feeds/sources/registry.move
+++ b/contracts/data-feeds/sources/registry.move
@@ -99,6 +99,11 @@ module data_feeds::registry {
     }
 
     #[event]
+    struct WriteSkippedFeedNotSet has drop, store {
+        feed_id: vector<u8>
+    }
+
+    #[event]
     struct OwnershipTransferRequested has drop, store {
         from: address,
         to: address
@@ -491,6 +496,8 @@ module data_feeds::registry {
         }
     }
 
+    // N.B. This FeedConfig type has a report field. These report fields are now being
+    // set to an empty vector<u8> on the on_report path(s).
     #[view]
     public fun get_feeds(): vector<FeedConfig> acquires Registry {
         let registry = borrow_global<Registry>(get_state_addr());
@@ -600,7 +607,11 @@ module data_feeds::registry {
     fun perform_update(
         registry: &mut Registry, feed_id: vector<u8>, report_data: vector<u8>
     ) {
-        if (!simple_map::contains_key(&registry.feeds, &feed_id)) { return };
+        if (!simple_map::contains_key(&registry.feeds, &feed_id)) {
+            event::emit(WriteSkippedFeedNotSet { feed_id });
+
+            return
+        };
 
         let feed = simple_map::borrow_mut(&mut registry.feeds, &feed_id);
 
--- a/contracts/data-feeds/sources/registry.move
+++ b/contracts/data-feeds/sources/registry.move
@@ -35,7 +35,7 @@ module data_feeds::registry {
         observation_timestamp: u256
     }
 
-    struct Benchmark has store, drop {
+    struct Benchmark has store, drop, copy {
         benchmark: u256,
         observation_timestamp: u256
     }
@@ -215,8 +215,7 @@ module data_feeds::registry {
         );
 
         let registry = borrow_global_mut<Registry>(get_state_addr());
-        let extend_ref = &registry.extend_ref;
-        let object_signer = object::generate_signer_for_extending(extend_ref);
+        let object_signer = object::generate_signer_for_extending(&registry.extend_ref);
 
         // callback for on_report_secondary function
         let cb_secondary =
@@ -601,10 +600,10 @@ module data_feeds::registry {
     fun perform_update(
         registry: &mut Registry, feed_id: vector<u8>, report_data: vector<u8>
     ) {
-        assert!(
-            simple_map::contains_key(&registry.feeds, &feed_id),
-            error::invalid_argument(EFEED_NOT_CONFIGURED)
-        );
+        if (!simple_map::contains_key(&registry.feeds, &feed_id)) {
+            return
+        };
+
         let feed = simple_map::borrow_mut(&mut registry.feeds, &feed_id);
 
         let is_v03 = vector::length(&report_data) == 9 * 32; // 288 bytes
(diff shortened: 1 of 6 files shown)


mxiao-cll (approved):

karen-stepanyan (dismissed):

contracts/data-feeds/sources/registry.move
```diff
@@ -35,7 +35,7 @@ module data_feeds::registry {
         observation_timestamp: u256
     }
 
-    struct Benchmark has store, drop {
+    struct Benchmark has store, drop, copy {
```
alecgard, line 40: Is this needed?

contracts/data-feeds/sources/registry.move
```diff
@@ -601,10 +600,8 @@ module data_feeds::registry {
     fun perform_update(
         registry: &mut Registry, feed_id: vector<u8>, report_data: vector<u8>
     ) {
-        assert!(
-            simple_map::contains_key(&registry.feeds, &feed_id),
-            error::invalid_argument(EFEED_NOT_CONFIGURED)
-        );
+        if (!simple_map::contains_key(&registry.feeds, &feed_id)) { return };
```
alecgard: Can we emit an event for this case so we can detect it?

alecgard (commented):
Can we add a comment to functions that include reports in their response, marking that the report data is disabled.

Also recommend we use disabled or similar rather than deprecated, as deprecated implies that it is still working just not recommended.

contracts/data-feeds/sources/registry.move
cawthorne: It is either that, or we loop over and copy the vector or benchmarks just to emit the event. Without copy the value can only be used once.

contracts/data-feeds/sources/registry.move
```diff
@@ -35,7 +35,7 @@ module data_feeds::registry {
         observation_timestamp: u256
     }
 
-    struct Benchmark has store, drop {
+    struct Benchmark has store, drop, copy {
         benchmark: u256,
```
JohnChangUK, line 41: Would `value` be more clear?

contracts/data-feeds/sources/router.move
```diff
@@ -66,18 +73,22 @@ module data_feeds::router {
 
     public fun get_benchmarks(
         _authority: &signer, feed_ids: vector<vector<u8>>, _billing_data: vector<u8>
-    ): vector<Benchmark> acquires Router {
-        let _router = borrow_global<Router>(get_state_addr());
+    ): vector<Benchmark> {
+        let benchmarks = registry::get_benchmarks_unchecked(feed_ids);
```
JohnChangUK, line 79: Is this intended, i.e. anyone can call this?

JohnChangUK (dismissed):

JohnChangUK:
Has the upgrade been tested (i.e. on devnet/testnet)? From previous version to current version?

The changes made should be backward compat, but best to verify.

contracts/data-feeds/sources/router.move
cawthorne: Anyone can call this function yep!

contracts/data-feeds/sources/registry.move
cawthorne: Probably yes, but we can't change it now as it is onchain!

cawthorne:
> Has the upgrade been tested (i.e. on devnet/testnet)? From previous version to current version?
> 
> The changes made should be backward compat, but best to verify.

This specific upgrade has only been tested on a local testnet.

We will only merge this PR in, as it is deployed in Production.

alecgard (approved):

contracts/data-feeds/sources/registry.move
```diff
@@ -645,14 +657,18 @@ module data_feeds::registry {
...
+            feed.report = vector::empty<u8>();
+        };
 
         event::emit(
             FeedUpdated {
                 feed_id,
                 timestamp: observation_timestamp,
                 benchmark: benchmark_price,
-                report: report_data
+                report: vector::empty<u8>()
```
JohnChangUK, line 671: Can just do `vector[]`
```suggestion
                report: vector[]
```

contracts/data-feeds/sources/registry.move
```diff
@@ -645,14 +657,18 @@ module data_feeds::registry {
 
         feed.observation_timestamp = observation_timestamp;
         feed.benchmark = benchmark_price;
-        feed.report = report_data;
+
+        if (vector::length(&feed.report) != 0) {
+            // Set to empty as value, as reports are deprecated.
+            feed.report = vector::empty<u8>();
```
JohnChangUK, line 663:
```suggestion
            feed.report = vector[];
```

contracts/data-feeds/sources/router.move
```diff
@@ -69,15 +76,21 @@ module data_feeds::router {
...
 
-        registry::get_benchmarks_unchecked(feed_ids)
+        let benchmarks = registry::get_benchmarks_unchecked(feed_ids);
+
+        event::emit(FeedRead { feed_ids: feed_ids, benchmarks: benchmarks });
+
+        benchmarks
     }
 
+    // This function has been disabled
```
JohnChangUK, line 86: Is this deprecated? If so can add `#[deprecated]` on top to make it clear.

contracts/data-feeds/sources/router.move
```diff
@@ -69,15 +76,21 @@ module data_feeds::router {
...
     public fun get_reports(
```
JohnChangUK, line 87:
```suggestion
    #[deprecated]
    public fun get_reports(
```

contracts/data-feeds/sources/router.move
```diff
@@ -69,15 +76,21 @@ module data_feeds::router {
     ): vector<Benchmark> acquires Router {
         let _router = borrow_global<Router>(get_state_addr());
```
JohnChangUK, line 77: We can remove this if not used? And the `acquires Router`?

JohnChangUK (approved):

PR #220, opened 2025-06-12
contracts/ccip/ccip: validate account address lengths

cfal:


contracts/ccip/ccip/sources/client.move
```diff
@@ -41,19 +43,35 @@ module ccip::client {
...
-            while (token_receiver.length() < 32) {
-                token_receiver.push_back(0);
-            };
-            token_receiver.reverse();
-        };
-        assert!(token_receiver.length() == 32, E_INVALID_SVM_TOKEN_RECEIVER_LENGTH);
+        pad_svm_address(&mut token_receiver);
+        assert!(
+            token_receiver.length() == 32,
+            error::invalid_argument(E_INVALID_SVM_TOKEN_RECEIVER_LENGTH)
```
JohnChangUK, line 49: I prefer not to use these `error::()` wrappers as it makes writing test cases harder, as it will produce an erroneous error code which you have to convert from decimal to hex.

JohnChangUK (approved):

contracts/ccip/ccip/sources/client.move
cfal: we are consistently using them across CCIP though

JohnChangUK: I see, we can keep this for now

PR #225, opened 2025-06-13
Separate register pool and setting pool in token_admin_registry.move

JohnChangUK:
## Desc
- Separate register pool and setting pool in token_admin_registry.move
- Removed Token Registrar logic as token owner or CCIP owner must call `set_pool`
- Removed `@token_pool_administrator` as we pass this in `token_admin_registry::set_pool`

## Flow
1. Create Pool
2. Propose admin for token
3. Accept admin
4. `set_pool`
    - Only callable by token admin


contracts/ccip/ccip/sources/token_admin_registry.move
```diff
@@ -522,51 +430,67 @@ module ccip::token_admin_registry {
...
+        let caller_addr = signer::address_of(caller);
 
-        event::emit(
-            TokenRegistrarSet {
-                local_token,
-                previous_token_registrar,
-                new_token_registrar: token_registrar
-            }
+        assert!(
+            object::owns(metadata, caller_addr) || caller_addr == auth::owner(),
```
JohnChangUK: Using `owns` to check each level of the object hierarchy.

cfal (dismissed):

contracts/ccip/ccip/sources/token_admin_registry.move
```diff
@@ -522,51 +433,104 @@ module ccip::token_admin_registry {
...
-    }
 
-    #[view]
-    public fun get_token_registrar(local_token: address): address acquires TokenAdminRegistryState {
+        assert!(administrator != @0x0, error::invalid_argument(E_ZERO_ADDRESS));
+
         let state = borrow_state_mut();
-        *state.token_registrars.borrow(&local_token)
+        assert!(
+            !state.token_configs.contains(&local_token),
```
friedemannf: This should instead check that the admin role hasn't been accepted yet. Right now we could only call this once, since a token_config would immediately be added.
This is to allow for multiple `propose_administrator` calls in case a wrong administrator has been proposed

friedemannf (changes requested):

contracts/ccip/ccip/sources/token_admin_registry.move
JohnChangUK: Good spot, amended so `propose_administrator` can be called multiple times, asserting that the current admin has not been set.

friedemannf (approved):

PR #227, opened 2025-06-15
Migrate SmartTable to BigOrderedMap

JohnChangUK:
## Desc
Migrate SmartTable to BigOrderedMap

## Limitations
`BigOrderedMap`
- When using `borrow_mut`, is either key or value has a variable size, `borrow_mut` reverts.
- From docs: https://github.com/aptos-labs/aptos-core/blob/96fa267/aptos-move/framework/aptos-framework/sources/datastructures/big_ordered_map.move#L450

"In case of variable size, use either `borrow`, `copy` then `upsert`, or `remove` and `add` instead of mutable borrow."

### DIFF
--- a/contracts/ccip/ccip/sources/fee_quoter.move
+++ b/contracts/ccip/ccip/sources/fee_quoter.move
@@ -10,7 +10,7 @@ module ccip::fee_quoter {
     use std::option;
     use std::signer;
     use std::string::{Self, String};
-    use std::smart_table::{Self, SmartTable};
+    use std::big_ordered_map::{Self, BigOrderedMap};
     use std::timestamp;
 
     use ccip::auth;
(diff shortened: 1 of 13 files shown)


contracts/ccip/ccip/sources/fee_quoter.move
```diff
@@ -461,16 +476,18 @@ module ccip::fee_quoter {
...
-    ) acquires FeeQuoterState {
+    ) acquires FeeQuoterState, FeeQuoterEvents {
         auth::assert_only_owner(signer::address_of(caller));
 
         let state = borrow_state_mut();
 
-        if (!state.token_transfer_fee_configs.contains(dest_chain_selector)) {
-            state.token_transfer_fee_configs.add(dest_chain_selector, smart_table::new());
-        };
         let token_transfer_fee_configs =
```
RensR: nit: current_token_transfer_fee_configs

contracts/ccip/ccip_offramp/sources/offramp.move
```diff
@@ -484,8 +487,13 @@ module ccip_offramp::offramp {
         // Execute the message
         execute_single_message(state, message, &execution_report.offchain_token_data);
 
-        // Since Aptos only supports success of reverts, when it reaches this it has succeeded.
-        *execution_state_ref = EXECUTION_STATE_SUCCESS;
+        // Since Aptos only supports success or reverts, when it reaches this it has succeeded.
+        // Use remove and add pattern due to variable size struct
+        // https://github.com/aptos-labs/aptos-core/blob/96fa267/aptos-move/framework/aptos-framework/sources/datastructures/big_ordered_map.move#L450
+        let source_chain_execution_states =
+            state.execution_states.remove(&source_chain_selector);
```
RensR: This feels extremely error prone. This exec state object is security critical and removing and re-adding it like this feels wrong to me. Is this really the only way? Can't we borrow and upsert like the other places?

RensR:
I really dislike the new pattern as it's incredibly easy to make mistakes but I guess there's no way around it

cfal:
i would lean towards not merging this in if all the SmartTables are internal. wdyt?

PR #243, opened 2025-06-21
Add pool address as identifier to events

JohnChangUK:
## Desc
Add pool address as identifier to events


cfal (changes requested):
let's discuss, i am not convinced this is the right fix.

our observability reads these events and expect it to be for the token pool at this package's address. breaking that invariant would mean a lot of extra tooling/offchain work (ie. comparing `address` only for Aptos, and two separate addresses needing to be tracked, one for the ccip_token_pool deployment, and the other being the real token pool deployment)

PR #245, opened 2025-06-23
Remove V2 events

JohnChangUK:


cfal (changes requested):
please also remove from `managed_token`

RensR (dismissed):

RensR:
Do we need to keep it in the dummy receiver? It's the only place in the ccip package that still has it.

MCMS also still has both

cfal:
> Do we need to keep it in the dummy receiver? It's the only place in the ccip package that still has it.
> 
> MCMS also still has both

perhaps rm from dummy receiver as well, just for consistency @JohnChangUK 🙏 
i think mcms only has the new ones, so we should keep it? also not abuseable

JohnChangUK:
> > Do we need to keep it in the dummy receiver? It's the only place in the ccip package that still has it.
> > MCMS also still has both
> 
> perhaps rm from dummy receiver as well, just for consistency @JohnChangUK 🙏 i think mcms only has the new ones, so we should keep it? also not abuseable

Will move it from them too. We should keep them in MCMS as those are the only events being emitted, and acts as a source of truth.

cfal (approved):

PR #256, opened 2025-06-26
Add Object resource group identifier

JohnChangUK:
## Desc
Add Object resource group identifier to CCIP resources. 

Provides better gas optimizations, as resources are accessed more efficiently.

These resources are stored within objects, hence the change. Provides ability to upgrade the same object.


cfal (commented):
i'm also not familiar with this, is there a reference somewhere that mentions this as a gas optimization? also, would it be a lot of effort to run a gas simulation (perhaps on a unit test that would come across this) to show the differences and that this is indeed an improvement?

PR #264, opened 2025-06-30
More offchain fixes

JohnChangUK:


sdrug (approved):

relayer/chainreader/chainreader.go
```diff
@@ -52,6 +52,7 @@ type aptosChainReader struct {
 }
 
 var _ types.ContractTypeProvider = &aptosChainReader{}
+var _ types.ContractReader = &aptosChainReader{}
```
cfal, line 55: not necessary because of NewChainReader

relayer/chainreader/event_poller.go
```diff
@@ -122,19 +122,28 @@ eventLoop:
 			}
 
 			if len(batchRecords) > 0 {
-				if err := a.dbStore.InsertEvents(ctx, batchRecords); err != nil {
-					return fmt.Errorf("syncEvent: failed to insert batch of events: %w", err)
+				err := a.dbStore.InsertEvents(ctx, batchRecords)
```
cfal, line 125: - @sdrug what causes a failure such that we should have to retry each one again? what makes us think that doing it 1 more time will work?
- can we create an InsertEvent for a single insertion?
- InsertEvents should be made to return the list of records that failed along with the aggregated error string. then we don't need to try every record again, even those that were successful.

relayer/chainreader/tx_poller.go
```diff
@@ -267,13 +267,31 @@ func (a *aptosChainReader) syncTransmitterTxs(ctx context.Context, transmitter a
 		}
 
 		if len(records) > 0 {
+			// Try batch insert first
```
cfal, line 270: can we make a helper function on dbStore since this pattern is repeated from above, if it's necessary?

relayer/chainreader/utils/msghasher.go
```diff
@@ -231,6 +235,9 @@ func computeMetadataHash(
 }
 
 func encodeUint256(n *big.Int) []byte {
+	if n == nil {
```
cfal, line 238: this should instead return a `([]byte, error)` when it's nil, because that means the message is malformed. we shouldn't automatically infer it as 0, eg tokenAmount zero https://github.com/smartcontractkit/chainlink/blob/05f6cc4b7a99027ec95ac00e91f97bafc554b011/core/capabilities/ccip/ccipaptos/msghasher.go#L209

relayer/chainreader/event_poller.go
sdrug: We should undo this change. There is a confusion in what version they still inspect. We already [addressed](https://github.com/smartcontractkit/chainlink-aptos/pull/217/files#diff-58ba4698cfffe3f8bfff783918183e5027c1b0c9697920ea3724407817b3d4f8) it.

relayer/chainreader/chainreader.go
JohnChangUK: I see, can remove this then

relayer/chainreader/tx_poller.go
JohnChangUK: Reverting this as discussed

relayer/chainreader/utils/msghasher.go
JohnChangUK: Agreed, will amend

PR #269, opened 2025-07-02
Add more error checks bcs ser/deser

JohnChangUK:


relayer/txm/bcs_util.go
```diff
@@ -221,7 +221,7 @@ func serializeArg(argVal any, argType aptos.TypeTag, serializer *bcs.Serializer)
 			return nil
 		}
 	case aptos.TypeTagU64:
-		if v, ok := argVal.(uint64); ok {
+		if v, ok := argVal.(uint64); ok && v == uint64(uint64(v)) {
```
cfal: is this correct? why do we need to `uint64(uint64(v))`, which is already converted to `uint64`?

relayer/txm/bcs_util.go
```diff
@@ -248,8 +248,10 @@ func serializeArg(argVal any, argType aptos.TypeTag, serializer *bcs.Serializer)
 		}
 		if v, ok := argVal.(uint64); ok {
 			b := big.NewInt(0).SetUint64(v)
-			serializer.U128(*b)
-			return nil
+			if v == b.Uint64() {
```
cfal: same question here and below

cfal (approved):

friedemannf (approved):

PR #282, opened 2025-07-23
Logpoller

sdrug:


relayer/chainreader/chainreader.go
```diff
@@ -476,11 +456,6 @@ func (a *aptosChainReader) QueryKey(ctx context.Context, contract types.BoundCon
 
 	eventHandle := address.String() + "::" + eventModuleName + "::" + eventConfig.EventHandleStructName
 
-	// Always sync event to get the latest data
-	if err := a.syncEvent(ctx, address, eventConfig, eventModuleName); err != nil {
-		return nil, fmt.Errorf("syncEvent error: %w", err)
-	}
```
faisal-chainlink, line 482: Does removing this mean we are now always dependent on the DB to get the latest available data?

sdrug: Yes, we now fully rely on the logpoller to populate the database. The idea is to allow the plugin to potentially iterate a few times before fetching the latest data, rather than risk blocking the call with a faulty RPC request. It's still experimental though.

cfal: even previously, we were always dependent on the DB, just that we would sync first.
we can re-add syncEvent, but we should use TryLock instead of locking on the mutex to prevent blocking the gRPC request.. IF we want to add it back in at all, which i'm not sure we should. even with try locking, RPC requests could be slow.

relayer/cache/cache.go
```diff
@@ -0,0 +1,106 @@
+package cache
```
cfal: i think we should avoid writing our own cache if possible. can you please check if go-cache fits our usecase? it is already imported in core, and indirectly in our integration-tests. https://github.com/smartcontractkit/chainlink/blob/55e7c8b5055c975665a59199d5eda9fa21801a07/core/services/ocr2/plugins/ccip/internal/cache/commit_roots.go#L9

relayer/chainreader/db/db.go
```diff
@@ -30,55 +30,6 @@ func NewDBStore(ds sqlutil.DataSource, logger logger.Logger) *DBStore {
 	}
 }
 
-func (s *DBStore) EnsureSchema(ctx context.Context) error {
```
rodrigombsoares, line 33: Was this unused already?

rodrigombsoares (approved):

PR #285, opened 2025-08-11
Regulated token

JohnChangUK:
## Summary
Regulated token with features:
  - Access control functionality - Responsible for configuring roles, RBAC system
  - Regulated token - main module

## 🌉 Bridge Functions
**Problem:** Dynamic dispatch conflicts cause CCIP token pool operations to fail with `fungible_asset::deposit()`/`withdraw()`.

**Solution:** Added bridge functions that bypass dispatch conflicts:
```move
public fun bridge_mint(caller: &signer, to: address, amount: u64): FungibleAsset
public fun bridge_burn(caller: &signer, from: address, fa: FungibleAsset)
```
- Use `MintRef`/`BurnRef` directly to avoid store operations
- Require `BRIDGE_MINTER_OR_BURNER_ROLE` authorization
- Maintain all security checks (pause/freeze protection)

## 🔄 Frozen Fund Recovery
**New Capability:** Transfer frozen funds to designated addresses (previously could only burn).

```move
public entry fun recover_frozen_funds(caller: &signer, from: address, to: address)
public entry fun batch_recover_frozen_funds(caller: &signer, accounts: vector<address>, to: address)
```


JohnChangUK:
Merged Regulated Token Pool PR into this https://github.com/smartcontractkit/chainlink-aptos/pull/291

contracts/regulated_token/sources/regulated_token.move
```diff
@@ -0,0 +1,1158 @@
...
+    enum Role has copy, drop, store {
+        PAUSER_ROLE(u8),
+        UNPAUSER_ROLE(u8),
+        FREEZER_ROLE(u8),
+        UNFREEZER_ROLE(u8),
+        MINTER_ROLE(u8),
+        BURNER_ROLE(u8),
+        BRIDGE_MINTER_OR_BURNER_ROLE(u8),
+        RECOVERY_ROLE(u8)
+    }
```
gregnazario: This does serialize down to a u8, so you don't need the u8, and the u8 inside, unless there's something specific you're using it for.

But, then at that point you might as well just use a u8, and the constants directly.

contracts/regulated_token/sources/regulated_token.move
```diff
@@ -0,0 +1,1158 @@
...
+
+    #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+    struct TokenStateDeployment has key {
+        extend_ref: ExtendRef,
+        transfer_ref: ObjectTransferRef,
+        paused: bool,
+        frozen_accounts: BigOrderedMap<address, bool>,
```
gregnazario, line 54: This should be fine, but I would probably opt to use an enum that contains the map for these, in the event you need to change / scale the frozen accounts list, and something changes along the way.

contracts/regulated_token/sources/regulated_token.move
```diff
@@ -0,0 +1,1158 @@
...
+    #[event]
+    struct AccountFrozen has drop, store {
+        freezer: address,
+        account: address
+    }
+
+    #[event]
+    struct AccountUnfrozen has drop, store {
+        freezer: address,
+        account: address
```
gregnazario: Depending if this is a factory of tokens, you may want to include the FA's address for indexing ease.

contracts/regulated_token/sources/regulated_token.move
```diff
@@ -0,0 +1,1158 @@
...
+    const E_ONLY_MINTER_OR_BRIDGE: u64 = 4;
+    const E_INVALID_ASSET: u64 = 5;
+    const E_ZERO_ADDRESS_NOT_ALLOWED: u64 = 6;
+    const E_CANNOT_TRANSFER_TO_REGULATED_TOKEN: u64 = 7;
+    const E_PAUSED: u64 = 8;
+    const E_ACCOUNT_FROZEN: u64 = 9;
+    const E_INVALID_ROLE_NUMBER: u64 = 10;
+    const E_INVALID_STORE: u64 = 11;
+    const E_TOKEN_ALREADY_INITIALIZED: u64 = 12;
+    const E_TOKEN_STATE_DEPLOYMENT_ALREADY_INITIALIZED: u64 = 13;
```
gregnazario: For errors, you should really put a `///` doc comment that gives a human readable error message.

contracts/regulated_token/sources/regulated_token.move
```diff
@@ -0,0 +1,1158 @@
...
+
+    #[view]
+    public fun type_and_version(): String {
+        string::utf8(b"RegulatedToken 1.0.0")
```
gregnazario, line 191: For your own purposes and ease of updating, would be good to make these two constants at the top and then use the constant here.

contracts/regulated_token/sources/regulated_token.move
```diff
@@ -0,0 +1,1158 @@
...
+            event::emit(
+                TokensRecovered {
+                    caller: signer::address_of(caller),
+                    token_metadata,
+                    from: @regulated_token,
+                    to,
+                    amount: balance
+                }
+            );
+        };
```
gregnazario: Probably want to move the transfer into the if statement, similar to the one below

contracts/regulated_token/sources/regulated_token.move
```diff
@@ -0,0 +1,1158 @@
...
+
+                event::emit(
+                    TokensRecovered {
+                        caller: signer::address_of(caller),
+                        token_metadata,
+                        from,
+                        to,
+                        amount: balance
+                    }
+                );
```
gregnazario: This blob could probably be a separate function `recover_funds` since you do it in 3 places.

contracts/regulated_token/sources/regulated_token.move
```diff
@@ -0,0 +1,1158 @@
...
+    /// Only callable by accounts with RECOVERY_ROLE.
+    public entry fun recover_frozen_funds(
+        caller: &signer, from: address, to: address
+    ) acquires TokenMetadataRefs, TokenState {
+        assert_not_paused();
+        assert!(to != @0x0, E_ZERO_ADDRESS_NOT_ALLOWED);
+        assert!(
+            to != @regulated_token && to != token_state_address_internal(),
+            E_CANNOT_TRANSFER_TO_REGULATED_TOKEN
+        );
```
gregnazario: Similarly the check for where to send recovered funds could be a function

contracts/regulated_token/sources/regulated_token.move
```diff
@@ -0,0 +1,1158 @@
...
+    public entry fun unpause(caller: &signer) acquires TokenState {
+        let state_obj = token_state_object_internal();
+        assert_unpauser(caller, state_obj);
+
+        let state = &mut TokenState[token_state_address_internal()];
+        if (state.paused) {
+            state.paused = false;
+            event::emit(Unpaused { unpauser: signer::address_of(caller) });
+        }
+    }
```
gregnazario: I guess these are to be idempotent rather than abort when it's already paused / unpaused?

contracts/regulated_token/sources/regulated_token.move
```diff
@@ -0,0 +1,1158 @@
...
+        assert_unfreezer(caller, token_state_object_internal());
+
+        primary_fungible_store::set_frozen_flag(
+            &borrow_token_metadata_refs().transfer_ref, account, false
+        );
+
+        TokenState[token_state_address_internal()].frozen_accounts.remove(&account);
+
+        event::emit(AccountUnfrozen { freezer: signer::address_of(caller), account });
+    }
```
gregnazario: Similarly, up to you if you want to emit a frozen / unfrozen event if the state hasn't changed.  Maybe you don't want to?

contracts/regulated_token/sources/regulated_token.move
```diff
@@ -0,0 +1,1158 @@
...
+
+        let burner = signer::address_of(caller);
+        let state_obj = token_state_object_internal();
+
+        let (is_bridge_burner, _) = assert_burner_and_get_type(burner, state_obj);
+
+        let token_metadata = token_metadata_from_state_obj(state_obj);
+        if (primary_fungible_store::is_frozen(from, token_metadata)) {
+            let balance = primary_fungible_store::balance(from, token_metadata);
+            if (balance > 0) {
```
gregnazario: This could actually be a cool inline

```
inline fun do_if_nonzero_balance(process: |u8|) {
   let balance = ...
   if (balance > 0) {
      process(balance);
   }
}
```

contracts/regulated_token/sources/regulated_token.move
JohnChangUK: This package will only ever deploy one token, so don't think it's needed.

contracts/regulated_token/sources/regulated_token.move
JohnChangUK: We can get rid of the u8.

contracts/regulated_token/sources/regulated_token.move
JohnChangUK: Will add that in

contracts/regulated_token/sources/regulated_token.move
JohnChangUK: We have the same pattern like this in EVM, so rather keep it consistent.

contracts/regulated_token/sources/regulated_token.move
JohnChangUK: Yeah, we should check if the account exists in the first place. If not then we can skip event emission

contracts/regulated_token/sources/regulated_token.move
JohnChangUK: I think it's better to just revert if the state is paused instead :)

contracts/regulated_token/sources/regulated_token.move
JohnChangUK: Added!

contracts/regulated_token/sources/regulated_token.move
JohnChangUK: Refactored to always emit the event, as we always set the frozen status on the primary store level.

bindings/ccip_token_pools/regulated_token_pool/regulated_token_pool.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (c RegulatedTokenPoolContract) Address() aptos.AccountAddress {
+	return c.address
+}
+
+func (c RegulatedTokenPoolContract) RegulatedTokenPool() module_regulated_token_pool.RegulatedTokenPoolInterface {
+	return c.regulatedTokenPool
+}
+
+var FunctionInfo = bind.MustParseFunctionInfo(
```
friedemannf, line 35: Pls add this FunctionInfo to the global registry here: https://github.com/smartcontractkit/chainlink-aptos/blob/d001e8172f11d0963abf6180f16135229f1032b4/bindings/bindings.go#L23-L23

bindings/regulated_token/regulated_token.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (l RegulatedTokenContact) Address() aptos.AccountAddress {
+	return l.address
+}
+
+func (l RegulatedTokenContact) RegulatedToken() module_regulated_token.RegulatedTokenInterface {
+	return l.regulatedToken
+}
+
+var FunctionInfo = bind.MustParseFunctionInfo(
```
friedemannf, line 35: Same here: https://github.com/smartcontractkit/chainlink-aptos/blob/d001e8172f11d0963abf6180f16135229f1032b4/bindings/bindings.go#L23-L23

contracts/regulated_token/README.md
```diff
@@ -0,0 +1,274 @@
...
+
+# Example with actual token address:
+aptos move run \
+  --function-id 0x772225b9cc6f60891b1866c5c57f3f8d8fb236173ab12fbb296cd77cc5a2b7ae::regulated_token::initialize \
+  --args u8:0 string:"RegulatedToken" string:"TKN" u8:8 string:"https://regulatedtoken.com/icon.png" string:"RegulatedToken"
+```
+
+**Parameters:**
+
+- `max_supply`: Optional maximum supply (use `0` for unlimited, or any number (`1000000000`) for capped)
```
friedemannf: Since the supply is an `option::Option`, does `0` actually mean no cap or would it limit the max supply to `0` instead?

bindings/ccip_token_pools/regulated_token_pool/regulated_token_pool.go
```diff
@@ -0,0 +1,85 @@
...
+// The token pool's administrator will be set to the deployer's account address.
+func DeployToObject(
+	auth aptos.TransactionSigner,
+	client aptos.AptosRpcClient,
+	ccipAddress,
+	mcmsAddress,
+	ccipTokenPoolAddress,
+	localTokenAddress aptos.AccountAddress,
+) (aptos.AccountAddress, *api.PendingTransaction, RegulatedTokenPool, error) {
+	namedAddresses := map[string]aptos.AccountAddress{
```
friedemannf, line 74: This seems like it is missing the `admin` NA?

contracts/regulated_token/README.md
JohnChangUK: Changed this to be more clear, i.e. passing an empty vec for unlimited supply or with a limit.

bindings/ccip_token_pools/regulated_token_pool/regulated_token_pool.go
JohnChangUK: It is, added in!

bindings/ccip_token_pools/regulated_token_pool/regulated_token_pool.go
JohnChangUK: Added

friedemannf (approved):

PR #288, opened 2025-08-16
.github/workflows: add check-tidy

jmank88:
https://smartcontract-it.atlassian.net/browse/CRE-710

Related:
- https://github.com/smartcontractkit/chainlink-ton/pull/113
- https://github.com/smartcontractkit/chainlink-tron/pull/26
- https://github.com/smartcontractkit/chainlink-sui/pull/349


.github/workflows/aptos-relayer.yml
```diff
@@ -30,3 +30,29 @@ jobs:
...
+        run: |
+          go install github.com/jmank88/gomods@v0.1.6
+          gomods tidy
+          git add --all
+          git diff --minimal --cached --exit-code
+
+      - name: Ensure mockery has been run
+        run: |
+          grep -rl "^// Code generated by mockery" | grep .go$ | xargs -r rm
+          find . -type f -name .mockery.yaml -execdir mockery \;
```
jmank88: `mockery` is not installed, but nix appears to be using `go-mockery`. How can we leverage that for this check?

jmank88: @archseer what about this repo? nix does not appear to be used by any CI currently. Is there a way to ensure that I have the same version installed if I just invoke it directly?

jmank88: @archseer  bump

.github/workflows/aptos-relayer.yml
```diff
@@ -29,4 +29,29 @@ jobs:
         CLI_VERSION: 7.7.0
 
     - name: Run tests
-      run: cd relayer && go test -v -count=1 -p=1 -tags=integration ./...
+      run: cd relayer && go test -count=1 -tags=integration ./...
```
jmank88: `-p=1` slows things down to a single package at a time, and `-v` adds all the noise of the successful tests to the logstream, crowding out the failed test logs.

jmank88: Restored `-p=1` since we depend on that currently. Why are we limiting ourselves this way?

friedemannf (approved):

PR #302, opened 2025-09-15
integration-tests/deploy: make container names unique

jmank88:
Unique container names in order to allow testing packages concurrently.

### DIFF
--- a/integration-tests/deploy/config.go
+++ b/integration-tests/deploy/config.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"testing"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


integration-tests/deploy/config.go
```diff
@@ -214,6 +215,9 @@ func ValidateGeth() (*GethConfig, error) {
 func ValidateDevnet() (*DevnetConfig, error) {
 	var errs []string
 	name := "chainlink-aptos.devnet"
+	if testing.Testing() {
```
jmank88, line 218: I suppose it doesn't need to be conditional on testing, though this is sufficient.

PR #304, opened 2025-09-15
bindgen: parse constants

friedemannf:
This adds parsing of constants to the generated bindings.
Currently supported are immutable types, such as integers but not vectors/strings.


cmd/bindgen/parse/parse.go
```diff
@@ -37,6 +37,11 @@ type ExternalStruct struct {
 	Module  string
 	Name    string
 }
```
rodrigombsoares, line 39: nit: line break

JohnChangUK (approved):

rodrigombsoares (approved):

PR #313, opened 2025-10-13
Dynamic dispatch support for CCIP Receiver contracts

JohnChangUK:
## Desc
Dynamic dispatch support for CCIP Receiver contracts


contracts/managed_token/sources/managed_token.move
```diff
@@ -336,6 +336,35 @@ module managed_token::managed_token {
         );
     }
 
+    public fun bridge_transfer(
```
JohnChangUK, line 339: The token pool could just call `burn` and then `mint`, this is more of a convenience function. I.e. just calling a single function from the receiver contract.

contracts/regulated_token/sources/regulated_token.move
```diff
@@ -592,6 +599,28 @@ module regulated_token::regulated_token {
         }
     }
 
+    public fun bridge_transfer(
```
JohnChangUK, line 602: The token pool could just call burn and then mint, this is more of a convenience function. I.e. just calling a single function from the receiver contract.

PR #314, opened 2025-10-14
Add out-of-order Feed Update Unit Test

cawthorne:
Unit test to ensure out of order reports are not accepted by the Data Feeds Registry (DF-21723)

### DIFF
--- a/contracts/data-feeds/sources/registry.move
+++ b/contracts/data-feeds/sources/registry.move
@@ -1473,4 +1473,3 @@ module data_feeds::registry {
         );
     }
 }
-
--- a/contracts/data-feeds/sources/registry.move
+++ b/contracts/data-feeds/sources/registry.move
@@ -1429,21 +1429,18 @@ module data_feeds::registry {
         let older_timestamp = 0x1f4; // 500 in decimal
         let older_benchmark = 0x1388; // 5000 in decimal
 
-        // Verify that StaleReport event was NOT emitted after the first (successful) update
-        assert!(
-            !event::was_event_emitted(
-                &StaleReport {
-                    feed_id,
-                    latest_timestamp: newer_timestamp,
-                    report_timestamp: older_timestamp
-                }
-            ),
-            10
-        );
+        let stale_report_event =
+            &StaleReport {
+                feed_id,
+                latest_timestamp: newer_timestamp,
+                report_timestamp: older_timestamp
+            };
 
-        registry = borrow_global_mut<Registry>(get_state_addr());
+        // Verify that StaleReport event was NOT emitted after the first (successful) update
+        assert!(!event::was_event_emitted(stale_report_event), 10);
 
         // Perform the update with the older report
+        let registry = borrow_global_mut<Registry>(get_state_addr());
         perform_update(registry, feed_id, older_report_data);
 
         // Verify that the value did NOT update (should still be the newer values)
@@ -1454,16 +1451,7 @@ module data_feeds::registry {
         assert!(benchmark.observation_timestamp == newer_timestamp, 6); // Should still be newer timestamp
 
         // Verify that StaleReport event was emitted for the out-of-order update
-        assert!(
-            event::was_event_emitted(
-                &StaleReport {
-                    feed_id,
-                    latest_timestamp: newer_timestamp,
-                    report_timestamp: older_timestamp
-                }
-            ),
-            8
-        );
+        assert!(event::was_event_emitted(stale_report_event), 8);
 
         // Verify that FeedUpdated event was NOT emitted for the stale report
         assert!(
@@ -1479,3 +1467,4 @@ module data_feeds::registry {
         );
     }
 }
+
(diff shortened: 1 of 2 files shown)


contracts/data-feeds/sources/registry.move
```diff
@@ -1365,4 +1365,117 @@ module data_feeds::registry {
...
+        assert!(
+            !event::was_event_emitted(
+                &StaleReport {
+                    feed_id,
+                    latest_timestamp: newer_timestamp,
+                    report_timestamp: older_timestamp
+                }
+            ),
+            10
+        );
```
cll-gg, line 1446: Nitpick: I'd move this to directly after checking that the `FeedUpdated` event was emitted

contracts/data-feeds/sources/registry.move
```diff
@@ -1365,4 +1365,117 @@ module data_feeds::registry {
...
+
+        registry = borrow_global_mut<Registry>(get_state_addr());
```
cll-gg: No need to do this again, I think?

contracts/data-feeds/sources/registry.move
```diff
@@ -1365,4 +1365,117 @@ module data_feeds::registry {
...
+        assert!(benchmark.observation_timestamp == newer_timestamp, 6); // Should still be newer timestamp
+
+        // Verify that StaleReport event was emitted for the out-of-order update
+        assert!(
+            event::was_event_emitted(
+                &StaleReport {
+                    feed_id,
+                    latest_timestamp: newer_timestamp,
+                    report_timestamp: older_timestamp
+                }
```
cll-gg: Would it be possible/nice to extract the StaleReport event and the FeedUpdated events into a reusable variable? Then we improve the test's resilience. 

(i.e. if someone updates `report` on line 1475 to be `vector[1]` then the test will still succeed but it won't actually test what we want anymore.)

contracts/data-feeds/sources/registry.move
cll-gg: Is not directly possible because we use the `older_timestamp` here

contracts/data-feeds/sources/registry.move
cll-gg: Actually it is

contracts/data-feeds/sources/registry.move
cll-gg: Extracted the StaleReport event. The FeedUpdated events are not equal so no reason to extract

cll-gg (dismissed):

contracts/data-feeds/sources/registry.move
```diff
@@ -1365,4 +1365,105 @@ module data_feeds::registry {
...
+
+        // Verify that the value did NOT update (should still be the newer values)
+        benchmarks = get_benchmarks(owner, vector[feed_id]);
+        assert!(vector::length(&benchmarks) == 1, 4);
+        benchmark = vector::borrow(&benchmarks, 0);
+        assert!(benchmark.benchmark == newer_benchmark, 5); // Should still be newer value
+        assert!(benchmark.observation_timestamp == newer_timestamp, 6); // Should still be newer timestamp
+
+        // Verify that StaleReport event was emitted for the out-of-order update
+        assert!(event::was_event_emitted(stale_report_event), 8);
```
JohnChangUK: We already assert this above. I assume we want this to be emitted twice. You should call `emitted_events()` and check the count and make sure they are what you expect.

cll-gg: You mean in [line 1440](https://github.com/smartcontractkit/chainlink-aptos/pull/314/files#diff-c2e2a78d782773cbd73c76e397f6634d35a88a87d1a4e2d0db82e04ee18be099R1440)? There we're checking that the StaleReport event has _not_ been emitted. 

We're checking here that the StaleReport event only gets emitted when a new report comes in with a lower observation timestamp.

JohnChangUK: I see. In that case you should also assert that no `StaleReport` events was emitted at all. i.e. getting `emitted_events()` and asserting 0

cll-gg: Good point, done!

cll-gg (dismissed):

contracts/data-feeds/sources/registry.move
```diff
@@ -1365,4 +1365,111 @@ module data_feeds::registry {
...
+                latest_timestamp: newer_timestamp,
+                report_timestamp: older_timestamp
+            };
+
+        // Verify that StaleReport event was NOT emitted after the first (successful) update
+        assert!(!event::was_event_emitted(stale_report_event), 5);
+
+        // Make sure no other events were emitted
+        assert!(
+            event::emitted_events<StaleReport>().length() == 0u64,
```
JohnChangUK:
```suggestion
            event::emitted_events<StaleReport>().length() == 0,
```

JohnChangUK (dismissed):

contracts/data-feeds/sources/registry.move
cll-gg: Done, thanks!

cll-gg (dismissed):

cll-gg (approved):

JohnChangUK (approved):

PR #338, opened 2026-02-13
relayer: defer getClient calls

jmank88:
It is critical that construction does not depend on external resources being online. This PR defers some client fetches that were being done in constructors, because the implementations block on getting chain IDs from the RPCs.

### DIFF
--- a/relayer/txm/txm.go
+++ b/relayer/txm/txm.go
@@ -42,21 +42,16 @@ type AptosTxm struct {
 	done          sync.WaitGroup
 	stop          chan struct{}
 
-	client aptos.AptosRpcClient
+	getClient func() (aptos.AptosRpcClient, error)
 }
 
 // TODO: Config input is not validated for sanity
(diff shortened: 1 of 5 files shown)


jmank88:
@amit-momin This is the sort of problem that I was looking for with our other recent issue. Although this log message was quite obvious, with a fast failure rather than a timeout.

relayer/txm/txm.go
```diff
@@ -469,8 +464,12 @@ func (a *AptosTxm) getTransactionAttempt(tx *AptosTx) uint64 {
 }
 
 func (a *AptosTxm) signAndBroadcast(tx *AptosTx) {
-	client := a.client
 	ctxLogger := GetContexedTxLogger(a.baseLogger, tx.ID, tx.Metadata)
+	client, err := a.getClient()
+	if err != nil {
+		ctxLogger.Error("Unable to sign and broadcast: failed to get client")
```
amit-momin: Think it could be helpful to print the err in this log and the one below

jmank88: Oops

amit-momin (approved):

cfal (approved):

PR #340, opened 2026-02-16
Adding aptos service implementation

yashnevatia:


relayer/chain/chain.go
```diff
@@ -73,10 +74,11 @@ var _ Chain = (*chain)(nil)
...
 
-	id   string
-	cfg  *config.TOMLConfig
-	lggr logger.Logger
-	ds   sqlutil.DataSource
+	id       string
+	cfg      *config.TOMLConfig
+	lggr     logger.Logger
+	ds       sqlutil.DataSource
+	keyStore loop.Keystore
```
yashnevatia, line 82: passing in keystore to use the highest balance transmitter account.
Check aptos_service.SubmitTransaction

relayer/aptos_service.go
```diff
@@ -0,0 +1,391 @@
...
+	}
+	s.logger.Infow("SubmitTransaction: accounts retrieved", "numAccounts", len(accounts))
+
+	// Find account with highest balance
+	publicKey, err := s.getAccountWithHighestBalance(ctx, accounts)
+	if err != nil {
+		s.logger.Errorw("SubmitTransaction: failed to get account with highest balance", "error", err)
+		return nil, fmt.Errorf("failed to determine account for SubmitTransaction: %w", err)
+	}
+	s.logger.Infow("SubmitTransaction: selected account", "publicKey", publicKey, "gasLimit", gasLimit.String())
```
Unheilbar: gasLimit can potentially be nil here?

relayer/aptos_service.go
```diff
@@ -0,0 +1,391 @@
...
+
+	result := make([]*commonaptos.Transaction, 0, len(txns))
+	for _, tx := range txns {
+		data, err := json.Marshal(tx.Inner)
+		if err != nil {
+			s.logger.Errorw("AccountTransactions: failed to marshal tx", "hash", string(tx.Hash()), "error", err)
+			return nil, fmt.Errorf("failed to marshal transaction data: %w", err)
+		}
+		v := tx.Version()
+		s := tx.Success()
```
Unheilbar: receiver's shadowing here, im curious if linter is okay with it? and if so, why?

Unheilbar (commented):
Ready to approve after a couple fixes

relayer/aptos_service.go
yashnevatia: good catch, removed

relayer/aptos_service.go
yashnevatia: good catch, i think because i am not using "s" inside the loop.
fixed.

relayer/txm/txm.go
```diff
@@ -180,11 +178,10 @@ func (a *AptosTxm) Enqueue(transactionID string, txMetadata *commontypes.TxMeta,
 		return fmt.Errorf("failed to parse contract address: %+w", err)
 	}
 
-	currentTimestamp := getTimestampSecs()
 	tx := &AptosTx{
 		ID:                             transactionID,
 		Metadata:                       txMetadata,
-		Timestamp:                      currentTimestamp,
+		Timestamp:                      getTimestampSecs(),
```
ilija42, line 184: why not inline this function?

relayer/txm/txm.go
```diff
@@ -197,35 +194,91 @@ func (a *AptosTxm) Enqueue(transactionID string, txMetadata *commontypes.TxMeta,
...
+// skipping the string-based function parsing and BCS serialisation of parameters.
+// The EntryFunction already contains the module, function name, type tags, and
+// pre-encoded BCS args.
+func (a *AptosTxm) EnqueueWithEntryFunction(transactionID string, txMetadata *commontypes.TxMeta, publicKey string, entryFunction *aptos.EntryFunction, simulateTx bool) error {
+	if entryFunction == nil {
+		return errors.New("entry function is required")
+	}
+
+	if transactionID == "" {
+		transactionID = uuid.New().String()
```
ilija42, line 210: Is this txId used anywhere, should this function return it?

relayer/txm/txm.go
yashnevatia: it was already available and used in 3 places now. So just reusing it.

relayer/txm/txm.go
yashnevatia: That txId is passed in from SubmitTransaction from aptos_service.
Having said that, you are right. If it is empty, we should return it.
https://github.com/smartcontractkit/chainlink-aptos/pull/340/commits/9bc322c07dfefe901be62c8afac3fa6ed5d02c06

ilija42 (approved):

Unheilbar (approved):

Fletch153 (approved):

friedemannf (approved):

PR #355, opened 2026-02-27
feat(onchain): Always execute OOO; Ignore allowOOOExecution in extraArgs

rodrigombsoares:
# Summary
Makes out-of-order execution always true on Aptos CCIP by ignoring the user’s value in extra args and overriding it to `true`. 
The contract still parses extra args as before but always treats/returns OOO as true and never reverts on the user’s choice, so existing users who omit the flag or set it to false are not broken.

# Features
- Fee quoter
  - Removed the three assertions that reverted when the user set `allow_out_of_order_execution == false`
  - After decoding extra args, the fee quoter always uses/returns OOO = `true`
  - Empty extra args defaults to `(default_tx_gas_limit, true)`
  - `enforce_out_of_order` remains in DestChainConfig for compatibility but is no longer used for any check.
- Onramp
  - Nonce is always set to 0 (Aptos CCIP always uses out-of-order execution).

# Testing
`test_out_of_order_execution_required` in `fee_quoter_chain_family.move` was updated to expect success when the user sends `create_extra_args(500000, false)`, and to assert that `process_message_args` returns `is_out_of_order_execution == true`.

To run tests:
`cd contracts/ccip/ccip`
`aptos move test --dev`

# Jira
https://smartcontract-it.atlassian.net/browse/NONEVM-3792


contracts/ccip/ccip_onramp/sources/onramp.move
```diff
@@ -611,13 +611,8 @@ module ccip_onramp::onramp {
             }
         );
 
-        let nonce =
-            if (is_out_of_order_execution) { 0 }
-            else {
-                nonce_manager::get_incremented_outbound_nonce(
-                    &state_signer, dest_chain_selector, sender
-                )
-            };
```
rodrigombsoares, line 620: This change here is optional, in theory `fee_quoter::process_message_args` will always return true for  `is_out_of_order_execution`.
But I feel like removing this gate is safer. WDYT @cfal @friedemannf?

contracts/ccip/ccip/sources/fee_quoter.move
```diff
@@ -916,8 +908,8 @@ module ccip::fee_quoter {
     ): (u256, bool) {
         let extra_args_len = extra_args.length();
         if (extra_args_len == 0) {
-            // If extra args are empty, generate default values.
-            (dest_chain_config.default_tx_gas_limit as u256, false)
+            // If extra args are empty, generate default values. Out-of-order is always true.
+            (dest_chain_config.default_tx_gas_limit as u256, true)
```
JohnChangUK: Can we have a const for `ALLOW_OOO_EXECUTION` and replace `true` with this

contracts/ccip/ccip_onramp/sources/onramp.move
RensR: I like this change! It in fact does mean you can fully remove the nonce manager as well

JohnChangUK (dismissed):

JohnChangUK (approved):

PR #380, opened 2026-03-10
Aptos local CRE minimal: expose Aptos service APIs and harden TXM submission

cawthorne:
## Summary
This PR upstreams the minimal Aptos work needed to run local CRE read-capability flows, while keeping the patch focused and reviewable.

### Relayer Aptos service surface
- Implemented Aptos service support on the relayer:
  - `Aptos()`
  - `AccountAPTBalance`
  - `LedgerVersion`
  - `View`
  - `TransactionByHash`
  - `AccountTransactions`
- Added Aptos type-tag/view payload conversion helpers and transaction conversion helpers.

### ChainReader head data support
- Added `GetLatestValueWithHeadData` for Aptos chainreader.
- Returns head metadata (height + timestamp) from node ledger info after read execution.

### Relayer/chain wiring cleanup
- Exposed chain keystore via `KeyStore()`.
- Refactored `GetChainInfo()` to use shared chain info helper.

### Dependency alignment
- Root module updated to a `chainlink-common` commit on upstream `main`.

## Notes / scope
- This PR is focused on local CRE + Aptos read-path enablement with minimal upstream scope.
- Historical `View` reads by explicit `ledger_version` are intentionally not included in this minimal patch and can be reintroduced in a follow-up once dependency stack alignment is complete.

## Validation
- `go test ./relayer/...`
- `cd integration-tests && go test ./smoke -run '^TestDoesNotExist$' -count=1` (compile validation)
- Full PR CI


cawthorne:
This PR depends on https://github.com/smartcontractkit/chainlink-common/pull/1885
And once https://github.com/smartcontractkit/chainlink-common/pull/1885 is merged in, we should update this PR to point to the upstreamed commit.

Fletch153 (dismissed):

relayer/relay.go
```diff
@@ -177,3 +211,232 @@ func (r *relayer) ListNodeStatuses(ctx context.Context, pageSize int32, pageToke
...
+		return nil, fmt.Errorf("failed to get client: %w", err)
+	}
+
+	payload, err := toSDKViewPayload(req.Payload)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []any
+	if ledgerVersion, ok := ledgerVersionFromViewRequest(req); ok {
```
ilija42: If we ignore invalid ledger version here is it possible for consensus reads to fail to reach consensus without knowing that the DON is not actually reading from a fixed ledgerVersion?

relayer/relay.go
```diff
@@ -177,3 +211,232 @@ func (r *relayer) ListNodeStatuses(ctx context.Context, pageSize int32, pageToke
...
+	data, err := json.Marshal(result)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal view response: %w", err)
+	}
+
+	return &typeaptos.ViewReply{Data: data}, nil
+}
+
+func ledgerVersionFromViewRequest(req typeaptos.ViewRequest) (uint64, bool) {
+	field := reflect.ValueOf(req).FieldByName("LedgerVersion")
```
ilija42: why do you need reflection here?

relayer/relay.go
```diff
@@ -177,3 +211,232 @@ func (r *relayer) ListNodeStatuses(ctx context.Context, pageSize int32, pageToke
...
+	client, err := r.chain.GetClient()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get client: %w", err)
+	}
+
+	tx, err := client.TransactionByHash(req.Hash)
+	if err != nil {
+		return nil, err
+	}
+	if tx == nil {
```
ilija42, line 254: Is this similar to how aptos handles txs that are not found? Should this be an error?

relayer/relay.go
```diff
@@ -177,3 +211,232 @@ func (r *relayer) ListNodeStatuses(ctx context.Context, pageSize int32, pageToke
...
+	}
+
+	txs, err := client.AccountTransactions(aptosdk.AccountAddress(req.Address), req.Start, req.Limit)
+	if err != nil {
+		return nil, err
+	}
+
+	out := &typeaptos.AccountTransactionsReply{Transactions: make([]*typeaptos.Transaction, 0, len(txs))}
+	for _, tx := range txs {
+		if tx == nil {
```
ilija42, line 284: Is it okay to just skip txs, similar question to txByHash

relayer/relay.go
cawthorne: It was a temporary shim while the chanlink-common patch was waiting to be upstreamed.

yashnevatia (changes requested):
Please rebase once this is merged in 
https://github.com/smartcontractkit/chainlink-aptos/pull/340

PR #388, opened 2026-03-13
feat: require explicit FromAddress in Workflow config (NONEVM-926)

Fletch153:
## Summary

- **Make `FromAddress` required** in TOML config when `[Workflow]` is set, so operators explicitly provide their on-chain account address instead of relying on implicit derivation from public key
- **Remove derivation fallback** in `getTransmitter()` (write_target) — `FromAddress` is guaranteed by config validation
- **Keep TXM derivation fallback** for CCIP compatibility (callers via `NewContractWriter` JSON path may not provide `FromAddress`)
- **Downgrade TXM derivation log** from warn-level comment to `Debugw` — it's normal behavior for CCIP callers

## Motivation

Aptos has native account abstraction: account addresses are permanent (`SHA3-256(original_pubkey || 0x00)`), but signing keys can be rotated. The codebase was implicitly deriving addresses from public keys, which:
- Fails for accounts with rotated authentication keys
- Caused a balance monitor bug
- Scattered address derivation across multiple layers

## Changes

| File | Change |
|------|--------|
| `relayer/config/config.go` | Uncomment `FromAddress` field, add validation requiring it when `[Workflow]` is set |
| `relayer/config/config_test.go` | Add tests: missing `FromAddress` fails, present passes, no `[Workflow]` passes |
| `relayer/write_target/aptos/write_target.go` | Wire `FromAddress` into chainwriter config, simplify `getTransmitter()`, remove unused `utils` import |
| `relayer/txm/txm.go` | Add `Debugw` log to derivation fallback (kept for CCIP compat) |

## What's NOT changed (and why)

- **`ChainWriterFunction.FromAddress`** — must stay optional for CCIP JSON deserialization compat (`relay.go`)
- **TXM `Enqueue()` fallback** — CCIP callers may not provide `fromAddress`
- **Balance monitor** — `KeyToAccountMapper` works correctly, monitors all keystore accounts
- **No startup sanity check** — warning on rotated keys every startup is noise

## Test plan

- [x] `go build ./...` compiles cleanly
- [x] `go test ./relayer/config/...` — validation tests pass (including new FromAddress tests)
- [x] `go test ./relayer/write_target/...` — write target tests pass
- [x] `go test ./relayer/txm/...` — TXM tests pass
- [x] `go test ./relayer/chainwriter/...` — chainwriter tests pass
- [x] TOML without `FromAddress` under `[Workflow]` fails validation with clear error
- [x] TOML without `[Workflow]` section still works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/relayer/config/config.go
+++ b/relayer/config/config.go
@@ -38,8 +38,8 @@ type ConfigSet struct { //nolint:revive
 
 type WorkflowConfig struct {
 	ForwarderAddress string
-	// FromAddress      string
-	PublicKey string
+	FromAddress      string
+	PublicKey        string
 }
 
 type Chain struct {
@@ -153,6 +153,13 @@ func (c *TOMLConfig) ValidateConfig() (err error) {
 		if err == nil && c.NetworkName != network.Name {
 			err = errors.Join(err, config.ErrInvalid{Name: "NetworkName", Value: c.NetworkName, Msg: fmt.Sprintf("does not match known network (%s) for chain ID", network.Name)})
 		}
+	}
+
+	if c.Workflow != nil && c.Workflow.FromAddress == "" {
+		err = errors.Join(err, config.ErrEmpty{
+			Name: "Workflow.FromAddress",
+			Msg:  "required - set it to your Aptos on-chain account address",
+		})
 	}
 
 	if len(c.Nodes) == 0 {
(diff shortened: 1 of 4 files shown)


Fletch153:
Duplicate — pushing changes to #387 instead.

PR #396, opened 2026-03-18
set NewNodeClientWithHttpClient params with cache

tt-cll:
Per AI:

Cause in chainlink-aptos (live)
In relayer/chain/chain.go, GetClient() creates a new aptos.NodeClient on every call and does not cache it:

Line 194: client, err = aptos.NewNodeClient(node.URL.String(), 0) is run each time.
Lines 222–226: A new rate-limited wrapper is created and returned; nothing is stored on the chain.
So every caller does this:

aptos_service.go – each View, AccountAPTBalance, TransactionByHash, AccountTransactions, etc. calls s.chain.GetClient().
txm – broadcast/confirm loops call getClient() when they need to talk to the chain.
logpoller – uses getClient() for polling.
balance monitor – uses NewClient: ch.GetClient.
relay.go – calls r.chain.GetClient().
write_target – calls chain.GetClient().
LatestHead – calls c.GetClient().
With aptos-go-sdk v1.12, each NewNodeClient gets its own http.Transport. So in production, every RPC path that calls GetClient() was creating a new client and a new transport. Under load that meant a new connection pool (and new connections) per call → port exhaustion.

So the problem is in chainlink-aptos’ usage: no client reuse, combined with v1.12’s per-client transport.

Change made in chainlink-aptos
relayer/chain/chain.go

Client cache

Added a cache keyed by node URL: clientCache map[string]aptos.AptosRpcClient with clientCacheMu sync.RWMutex.
GetClient() behavior

Try nodes in random order (unchanged).
Cache read: For each node URL, check the cache; if a client exists, call GetChainId() to validate it’s still good, then return that cached client.
Cache miss: Create aptos.NewNodeClient(...), validate chain ID, wrap in ratelimit.NewRateLimitedClient, store in the cache, and return it.
So in live environments you now have at most one client (and one Transport) per node URL per chain, reused by aptos_service, txm, logpoller, balance monitor, relay, write_target, and LatestHead. That removes the “new client every call” behavior that was driving port exhaustion with v1.12.

Summary
Layer	Role
aptos-go-sdk v1.12	Each NewNodeClient() uses its own http.Transport (TLS 1.2 change) → many transports if many clients are created.
chainlink-aptos (live)	Before: Every GetClient() created a new Node

### DIFF
--- a/relayer/chain/chain.go
+++ b/relayer/chain/chain.go
@@ -2,10 +2,13 @@ package chain
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"math/big"
 	"math/rand"
+	"net/http"
+	"net/http/cookiejar"
 	"strconv"
 	"sync"
 	"time"
@@ -86,6 +89,11 @@ type chain struct {
 	// every GetClient() call, which leads to port exhaustion under load.
 	clientCacheMu sync.RWMutex
 	clientCache   map[string]aptos.AptosRpcClient
+
+	// sharedTransport is used by all NodeClients so connections are pooled (one Transport
+	// per chain). TLS 1.2 and MaxIdleConnsPerHost avoid port exhaustion with SDK v1.12+.
+	sharedTransport     *http.Transport
+	sharedTransportOnce sync.Once
 
 	// Sub-services
 	txm            *txm.AptosTxm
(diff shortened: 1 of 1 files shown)


relayer/chain/chain.go
```diff
@@ -176,57 +192,116 @@ func (c *chain) KeyStore() loop.Keystore {
...
+// sharedHTTPTransport returns a Transport with TLS 1.2 and connection pool settings
+// so that all NodeClients for this chain share the same connection pool (avoids
+// port exhaustion with aptos-go-sdk v1.12+).
+func (c *chain) sharedHTTPTransport() *http.Transport {
+	c.sharedTransportOnce.Do(func() {
+		c.sharedTransport = &http.Transport{
+			TLSClientConfig: &tls.Config{
+				MinVersion: tls.VersionTLS12,
+			},
+			MaxIdleConns:        100,
```
tt-cll: default transport from golang for reference.
```
var DefaultTransport RoundTripper = &Transport{
	Proxy: ProxyFromEnvironment,
	DialContext: defaultTransportDialContext(&net.Dialer{
		Timeout:   30 * time.Second,
		KeepAlive: 30 * time.Second,
	}),
	ForceAttemptHTTP2:     true,
	MaxIdleConns:          100,
	IdleConnTimeout:       90 * time.Second,
	TLSHandshakeTimeout:   10 * time.Second,
	ExpectContinueTimeout: 1 * time.Second,
}
```

cfal (approved):

PR #429, opened 2026-04-13
chore(ci): bump deps, fixes

erikburt:
* Bumps many dependencies to their latest available version (many of which now use node24)
* Update inline shell scripts to use env vars
* Consolidate go to version `1.25.7`


chainchad (approved):
<3

timweri (approved):

amit-momin (approved):

PR #437, opened 2026-04-20
feat(relayer): transmitter address override for key rotation [PLEX-2632]

Fletch153:
## Summary

Aptos account addresses are derived from the public key at creation (`SHA3-256(pubkey || 0x00)`), but once the account's authentication key is rotated, the address stays fixed and can no longer be derived from the new key. This PR adds an optional per-key address override so the relayer points at the correct on-chain account after rotation.

- New `[Aptos.Transmitter.Overrides]` TOML section — map of keystore public key → on-chain account address. Absent / unlisted keys keep the existing derive-from-pubkey behavior.
- `transmitter.Config.ResolveAddress(pubKey)` returns override or derivation; validates TOML entries and rejects semantic duplicates (keys differing only in case, whitespace, or `0x` prefix).
- `AptosTxm.EnqueueWithEntryFunction` gains a `fromAddress aptos.AccountAddress` param; zero value preserves the existing derive fallback.
- `aptos_service.SubmitTransaction` and the balance monitor resolve the address per keystore pubkey through the resolver, so both submission and balance reporting target the rotated account.

TOML example:
```toml
[Aptos.Transmitter.Overrides]
"0x<keystore-pubkey>" = "0x<rotated-on-chain-address>"
```

Capabilities-side code is unchanged — the capability still submits `(ReceiverModuleID, EncodedPayload, GasConfig)` with no sender, and the relayer picks the address. For a full rotation, operators also need to update `p2pToTransmitterMap` at the capability registry (out of scope here).

## Test plan

- [x] `relayer/transmitter/config_test.go` — `ValidateConfig` / `ResolveAddress` across valid entries, malformed hex, wrong-length keys, invalid addresses, semantic duplicates, 0x-prefix + whitespace + case normalization, independent multi-key resolution.
- [x] `relayer/config/config_test.go` `TestTOMLConfig_Transmitter` — raw TOML decode path: override resolves, unlisted pubkey derives, malformed address/pubkey rejected at validation.
- [x] `relayer/txm/enqueue_entry_function_test.go` — non-zero `fromAddress` lands as `tx.FromAddress`; zero value falls back to derive.
- [x] `relayer/txm/txm_local_test.go` integration calls pass `fromAddress` explicitly to exercise the non-zero param path end-to-end against a live Aptos node (behind `//go:build integration`).
- [ ] Follow-up: dedicated rotation integration test against a live node (requires `0x1::account::rotate_authentication_key` proof-of-knowledge flow) — separate ticket.

## Related

- [PLEX-2632](https://smartcontract-it.atlassian.net/browse/PLEX-2632)

[PLEX-2632]: https://smartcontract-it.atlassian.net/browse/PLEX-2632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/relayer/aptos_service.go
+++ b/relayer/aptos_service.go
@@ -12,7 +12,6 @@ import (
 	"github.com/google/uuid"
 
 	"github.com/smartcontractkit/chainlink-aptos/relayer/chain"
-	"github.com/smartcontractkit/chainlink-aptos/relayer/utils"
 	"github.com/smartcontractkit/chainlink-common/pkg/logger"
 	commontypes "github.com/smartcontractkit/chainlink-common/pkg/types"
 	commonaptos "github.com/smartcontractkit/chainlink-common/pkg/types/chains/aptos"
@@ -218,6 +217,13 @@ func (s *aptosService) SubmitTransaction(ctx context.Context, req commonaptos.Su
 	}
 	s.logger.Infow("SubmitTransaction: selected account", "publicKey", publicKey)
 
+	fromAddress, err := s.chain.Config().Transmitter.ResolveAddress(publicKey)
+	if err != nil {
+		s.logger.Errorw("SubmitTransaction: failed to resolve sender address", "publicKey", publicKey, "error", err)
+		return nil, fmt.Errorf("failed to resolve sender address: %w", err)
+	}
+	s.logger.Debugw("SubmitTransaction: resolved sender address", "fromAddress", fromAddress.String())
+
 	txID := uuid.New().String()
 	s.logger.Infow("SubmitTransaction: enqueueing to TxManager", "txID", txID)
 	_, enqueueErr := s.chain.TxManager().EnqueueWithEntryFunction(
@@ -226,6 +232,7 @@ func (s *aptosService) SubmitTransaction(ctx context.Context, req commonaptos.Su
 			GasLimit: gasLimit,
 		},
 		publicKey,
+		fromAddress,
 		entryFn,
 		*s.chain.Config().AptosService.SimulateTx,
 		// TODO: add expected simulation failures to save gas on reported transmissions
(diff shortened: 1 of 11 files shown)


relayer/aptos_service.go
```diff
@@ -324,9 +331,9 @@ func (s *aptosService) getAccountWithHighestBalance(ctx context.Context, account
 	var foundAny bool
 
 	for _, account := range accounts {
```
yashnevatia: can we rename these to publicKeys please.
and can we call the addr var accountAddress please

relayer/aptos_service.go
```diff
@@ -218,6 +217,13 @@ func (s *aptosService) SubmitTransaction(ctx context.Context, req commonaptos.Su
 	}
 	s.logger.Infow("SubmitTransaction: selected account", "publicKey", publicKey)
 
+	fromAddress, err := s.chain.Config().Transmitter.ResolveAddress(publicKey)
```
yashnevatia: can we return this from the `getAccountWithHighestBalance` function instead of deriving it here separately.

relayer/aptos_service_test.go
```diff
@@ -97,9 +99,47 @@ func ptrUint64(v uint64) *uint64 {
...
+			Transmitter: &transmitter.Config{Overrides: map[string]string{pubKeyB: overrideAddr}},
+		},
+	}
+	svc := aptosService{
+		chain:  &testChain{client: client, config: cfg},
+		logger: logger.Test(t),
+	}
+
+	selected, _ := svc.getAccountWithHighestBalance(context.Background(), []string{pubKeyA, pubKeyB})
+	require.Equal(t, pubKeyB, selected, "picker must select the key whose resolved address holds the higher balance")
```
yashnevatia, line 137: return account here as well. and then can add require.Equal on overrideAddr as well. please call it overrideAcctAddr

relayer/txm/txm.go
```diff
@@ -197,11 +197,10 @@ func (a *AptosTxm) Enqueue(transactionID string, txMetadata *commontypes.TxMeta,
 	return a.enqueueTransaction(tx)
 }
 
-// EnqueueWithEntryFunction is like Enqueue but accepts a deserialized EntryFunction directly,
-// skipping the string-based function parsing and BCS serialisation of parameters.
-// The EntryFunction already contains the module, function name, type tags, and
```
yashnevatia, line 202: this part of the comment got deleted

yashnevatia (approved):

PR #447, opened 2026-04-29
fix(relayer): use Success flag to discriminate reverted txs (PLEX-2886)

Fletch153:
Closes [PLEX-2886](https://smartcontract-it.atlassian.net/browse/PLEX-2886) (L6).

## Summary

`aptos_service.SubmitTransaction` derived `TxStatus` from `VmStatus` emptiness:

```go
if txResult.VmStatus != "" {
    return commonaptos.TxReverted, nil
}
```

Aptos REST returns `vm_status: "Executed successfully"` on every committed non-aborted user transaction — see the official `aptos-go-sdk` test fixtures, e.g. https://github.com/aptos-labs/aptos-go-sdk/blob/main/api/transactions_test.go#L20-L22:

```json
{
  "success": true,
  "vm_status": "Executed successfully",
  ...
}
```

So a successful `forwarder::report` was returned to the capability layer as `TxReverted`. `WriteReport` recovered via the duplicate path, which depends on a 3-phase account-history scan to recover the original tx hash — an extra RPC roundtrip per success.

## Fix

Discriminate on the explicit `Success` boolean instead of inferring from `VmStatus`:

- Add `Success bool` to `txm.AptosTx` and `txm.TransactionResult`
- Populate it from `userTx.Success` alongside the existing `VmStatus` capture
- `aptos_service.SubmitTransaction` now branches on `!txResult.Success`

`VmStatus` stays as diagnostic context on the revert warn log.

## Test plan

- [x] `go build ./...`
- [x] `go test ./relayer/...`
- [ ] Validate against a live Aptos endpoint (committed tx with `success=true, vm_status="Executed successfully"` is no longer classified as `TxReverted`)

[PLEX-2886]: https://smartcontract-it.atlassian.net/browse/PLEX-2886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yashnevatia:
Had fixed that here https://github.com/smartcontractkit/chainlink-aptos/pull/443
but your solution is better

yashnevatia (dismissed):

yashnevatia (approved):

PR #449, opened 2026-04-30
Fix Aptos account transaction limit windows

cawthorne:
## Summary
- resolve `AccountTransactions(start=nil, limit=N)` into an explicit bounded window before calling the Aptos SDK
- return an empty result for zero-sequence accounts instead of issuing an invalid SDK request
- keep explicit-start account transaction requests unchanged
- lower expected empty-account transaction logs to debug for new/young transmitter accounts
- add relayer unit tests for young accounts, normal windows, zero sequence number, and explicit-start passthrough

## Root Cause
The Chainlink merge queue Aptos CRE test started requesting recent transmitter account transactions with `start=nil, limit=10`. That call shape was introduced by smartcontractkit/capabilities#574, specifically https://github.com/smartcontractkit/capabilities/commit/133e87d6fa49e380db7a4648587ef4dbd314e6b7, when the Aptos tx-search phase-1 probe changed from `start=nil, limit=nil` to a bounded latest-page request.

For limit-only account transaction requests, the Aptos SDK queries the latest transactions and then computes a previous-page cursor from the first returned transaction sequence number. For new accounts with fewer transactions than the requested limit, the earliest returned sequence can be `0`, so the SDK computes `0 - 1` and issues a REST request with `start=18446744073709551615`. The observed CI failure was the Aptos REST API rejecting that invalid cursor.

## Fix
This PR handles the limit-only request shape in the relayer by resolving it into a safe explicit latest transaction window before calling the Aptos SDK. For accounts with sequence number `0`, the relayer returns an empty transaction list directly. Requests that already provide an explicit `start` are still passed through unchanged.

## Testing
- `go test ./relayer -count=1`

Failing runs / consumers:
- https://github.com/smartcontractkit/chainlink/actions/runs/25151620823/job/73724531635
- https://chainlink-core.slack.com/archives/C043E82DUM8/p1777533874251039
- Chainlink consumer PR: https://github.com/smartcontractkit/chainlink/pull/22264

Aptos SDK issue filed:
https://github.com/aptos-labs/aptos-go-sdk/issues/213

### DIFF
--- a/relayer/aptos_service.go
+++ b/relayer/aptos_service.go
@@ -149,7 +149,7 @@ func (s *aptosService) AccountTransactions(ctx context.Context, req commonaptos.
 		return nil, err
 	}
 	if limit != nil && *limit == 0 {
-		s.logger.Infow("AccountTransactions: returning empty result", "address", sdkAddr.String())
+		s.logger.Debugw("AccountTransactions: returning empty result", "address", sdkAddr.String())
 		return &commonaptos.AccountTransactionsReply{}, nil
 	}
 
--- a/relayer/aptos_service.go
+++ b/relayer/aptos_service.go
@@ -193,6 +193,10 @@ func (s *aptosService) accountTransactionsWindow(
 		return start, limit, nil
 	}
 
+	// Fetch the sequence number to compute a safe window. A small TOCTOU race
+	// exists: new txns committed between Account() and AccountTransactions() may
+	// not appear in the result. That is acceptable — the caller can re-query.
+	// The goal is to prevent underflow, not to guarantee atomicity.
 	accountInfo, err := client.Account(address)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to get account info: %w", err)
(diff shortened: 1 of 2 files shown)


relayer/aptos_service.go
```diff
@@ -143,7 +143,17 @@ func (s *aptosService) AccountTransactions(ctx context.Context, req commonaptos.
...
 
 	sdkAddr := aptos_sdk.AccountAddress(req.Address[:])
-	txns, err := client.AccountTransactions(sdkAddr, req.Start, req.Limit)
+	start, limit, err := s.accountTransactionsWindow(client, sdkAddr, req.Start, req.Limit)
+	if err != nil {
+		s.logger.Errorw("AccountTransactions: failed to resolve transaction window", "address", sdkAddr.String(), "error", err)
+		return nil, err
+	}
+	if limit != nil && *limit == 0 {
+		s.logger.Infow("AccountTransactions: returning empty result", "address", sdkAddr.String())
```
cll-gg: No clue how often this will occur: is `info` the right level here?

cawthorne: Fair point, making it debug.

relayer/aptos_service.go
```diff
@@ -143,7 +143,17 @@ func (s *aptosService) AccountTransactions(ctx context.Context, req commonaptos.
 	}
 
 	sdkAddr := aptos_sdk.AccountAddress(req.Address[:])
-	txns, err := client.AccountTransactions(sdkAddr, req.Start, req.Limit)
+	start, limit, err := s.accountTransactionsWindow(client, sdkAddr, req.Start, req.Limit)
```
yashnevatia: instead of making 2 rpc calls per function call, can we please do the nil,nil call if the first one fails for whatever reason.

yashnevatia (approved):

JohnChangUK (approved):

PR #450, opened 2026-04-30
[audit-rebase] #437 — transmitter address override for key rotation

Fletch153:
Mirrors #437 (transmitter override for aptos key rotation, PLEX-2632) onto `bugfix/audit_db`. Origin: `866ccd8`.


Fletch153:
Closing — original PR #437 is being reverted on develop (#452). This rebased mirror is no longer in audit scope.

PR #479, opened 2026-06-03
ChainReader - Deserializer error handling

friedemannf:


relayer/chainreader/utils/deserializer.go
```diff
@@ -90,38 +108,68 @@ func DeserializeExecutionReport(data []byte) (*ExecutionReport, error) {
...
 	tokenAmountsLen := deserializer.Uleb128()
-	tokenAmounts := make([]Any2AptosTokenTransfer, tokenAmountsLen)
+	if err := deserializer.Error(); err != nil {
+		return nil, fmt.Errorf("failed to deserialize tokenAmounts length: %w", err)
+	}
+	var tokenAmounts []Any2AptosTokenTransfer
 
 	for i := uint32(0); i < tokenAmountsLen; i++ {
 		sourcePoolAddr := deserializer.ReadBytes()
+		if err := deserializer.Error(); err != nil {
```
FelixFan1992, line 143: i think these checks are not necessary?

we only need this check when reading a vector length and is going to allocate some memory. in ReadBytes or U64, etc, these read functions are all bounded and will return nil

JohnChangUK (approved):

PR #488, opened 2026-06-16
Add multinode support for aptos service

yashnevatia:
- Only touching aptos_service
- Not touching CCIP flow (no txm / chainreader / chainwriter)

### DIFF
--- a/relayer/config/config.go
+++ b/relayer/config/config.go
@@ -4,11 +4,13 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/aptos-labs/aptos-go-sdk"
 	"github.com/pelletier/go-toml/v2"
 
 	"github.com/smartcontractkit/chainlink-common/pkg/config"
+	mncfg "github.com/smartcontractkit/chainlink-framework/multinode/config"
 
 	"github.com/smartcontractkit/chainlink-aptos/relayer/logpoller"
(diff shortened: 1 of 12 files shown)


relayer/config/config.go
```diff
@@ -79,6 +84,55 @@ type TOMLConfig struct {
...
+// DefaultRequestTimeout bounds each individual Aptos RPC call when RequestTimeout is unset.
+const DefaultRequestTimeout = 30 * time.Second
+
+// SetMultiNodeDefaults fills any unset MultiNode field with an Aptos-appropriate default. The
+// framework's config accessors dereference these pointers directly, so every field consumed by
+// the node/multinode lifecycle must be non-nil. Tuned to Aptos's ~1-2s block time and its
+// single-finality model (a committed block is final: no reorgs).
+func (c *TOMLConfig) SetMultiNodeDefaults() {
+	m := &c.MultiNode.MultiNode
+	setDefault(&m.Enabled, true)
```
dhaidashenko, line 106: We might want to disable it by default, so that it does not affect CCIP setup.
Not sure if they want to have multinode enabled. We got a pushback on integrating it during initial development.

yashnevatia: hi, the way i have used it is that i dont us the multinode client for txm/chainreader/chainwriter.
only for aptos_service.go read methods.

dhaidashenko: But it'll still be running and producing requests/error/panics in the worst case.

PR #493, opened 2026-07-01
Codec Module

jmank88:
This PR creates a separate, top-level module `codec`, which contains the `relayer/codec` and `relayer/chainreader/loop` packages. This new module has ~100 less dependencies than the main module. The old packages are deprecated but remain compatible by forwarding to the new module.

Supports:
- https://github.com/smartcontractkit/chainlink/pull/23011

TODO:
- [x] Does CI run the unit tests in their new location?
- [ ] ~Split into two PRs to satisfy CI check for non-default branch commit hashes~ circle back to remove relative replaces


jmank88:
The 10 linter issues were pre-existing in code that was copied, and so can be ignored. The failure validates the matrix logic though 🙌

JohnChangUK (approved):

